PR #57, opened 2015-01-03
Add note on events to FAQ

dobrite:
No idea if the FAQ is the place for this, or even if it belongs, but I lost a few hours figuring this out and wanted to save others from the same fate. :grin:


staltz:
This piece of information will do fine in the FAQ for now. Thanks for finding the problem, solving it, and sharing the solution. I'll also look into incorporating `renderer.delegator.listenTo('mouseover');` et al to the framework.

PR #75, opened 2015-02-06
Add support for DocumentFragment as input for renderer

erykpiast:
Currently it's impossible to use Cycle.js with custom elements and Shadow DOM, because shadowRoot is DocumentFragment, not HTMLElement. As far as I see, virtual-dom package works fine with DocumentFragment too, so I see no reason to prevent usage of it in Cycle.


staltz:
Good discovery. Thanks for the PR and also the mocha test in it. :+1:

Issue #108, opened 2015-05-02
Stream can be better


pH200:
Let's start talking about Stream. This is long so TL;DR at the bottom.

First of all, why replicateAllInteraction$? You know it's unnecessary if everything was implemented just right. Because Stream can just be, like you said, InjectableObservable, without doing magic tricks just for interaction$. You want arguments for definitionFn observables. So you can replace them with subjects, except for interaction$ which is not actually an Observable. So you made InputProxy just for interaction$. No, it didn't have to be this way. You can inject normal objects and no tricks for interaction$. I'll show you later.

Second, sorry I was wrong about the timing that the observable returned by definitionFn gets subscribed. But, it's still not good the fact it gets subscribed by injection. Why inject() has to evaluate the observable if we can make it not to do so, like so many other [operators](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/observable.md) by default?

And that brings up another bad thing about the Stream right now: It appends the dispose method to the observable(returned by definitionFn). And as soon as you called inject, you were responsible to dispose that Stream(InjectableObservable). Because it fires up a subscription. This is bad. For example:

```js
var injectableObservable = createStream(_interval => _interval);
var interval = Rx.Observable.interval(500);

// This subscription will end eventually.
// You are free whether to dispose it or not.
var subscription = injectableObservable
  .take(3)
  .map(x => x * x)
  .subscribe(next => console.log(next));

// However, the hidden subscription which spawned by
// the injection would never stop.
// The interval will keep popping new value. It never ends.
injectableObservable.inject(interval);

// Unless you call dispose. This is just weird.
// Because you normally don't call dispose on an Observable.
// And this proves inject() is not the only point where mutation
// happens, if you don't forget to call dispose.
injectableObservable.dispose();
```

In Rx library, the only derivation of Observable which has dispose() in it is called Subject. However, it never fires up a new subscription when you call onNext from it. Again, it can be better.

There's more I want to say about the implementation of render and custom-elements. But for now I'll focus on the Stream issue.

So finally, this is my attempt at making Stream better. It doesn't fire up a new subscription when you called inject. It doesn't distinguish between the normal value and interaction$. It can inject normal objects not only just Observables and interaction$. Please go take a look, I'm very **confident** it will blow your mind.

https://github.com/pH200/cycle/blob/master/src/stream.js

[...]

Issue #110, opened 2015-05-04
Cannot catch events coming from custom element that patching itself


pH200:
I know you're busy working on nostream branch. But still, we should get this test passed for the upcoming merge.

test/browser/custom-elements.js

```js
it('should catch interaction events coming from custom element that patching itself', function (done) {
  // Make simple custom element
  Cycle.registerCustomElement('myelement', function (rootElem$) {
    // TODO these tests should not require shareReplay(1)! Remove and fix src/
    let vtree$ = Rx.Observable.merge(
      Rx.Observable.just(h('h3.myelementclass', 'foo')),
      Rx.Observable.just(h('button.myelementclass', 'bar')).delay(50)
    ).shareReplay(1);
    rootElem$.inject(vtree$);
    return {
      myevent$: Rx.Observable.just(123).delay(300)
    };
  });
  // Use the custom element
  let vtree$ = Rx.Observable.just(h('div.toplevel', [
    h('myelement.eventsource', {key: 1})
  ]));
  let rootElem$ = Cycle.createStream(function (vtree$) {
    return Cycle.render(vtree$, createRenderTarget());
  });
  rootElem$.interaction$.choose('.eventsource', 'myevent').subscribe(function (x) {
    assert.strictEqual(x.data, 123);
    rootElem$.dispose();
    done();
  });
  rootElem$.inject(vtree$);
  // Make assertions
  let myElement = document.querySelector('.myelementclass');
  assert.notStrictEqual(myElement, null);
  assert.notStrictEqual(typeof myElement, 'undefined');
  assert.strictEqual(myElement.tagName, 'H3');
});
```

Reference: https://gitter.im/staltz/cycle?at=554764d95b12f7f26b6aaaac

staltz:
Done

PR #111, opened 2015-05-05
nostream: Properly dispose resources after the custom-element has been destroyed

pH200:
Reference: #108

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   ],
   "main": "lib/cycle.js",
   "dependencies": {
+    "es6-map": "^0.1.1",
     "rx": "2.5.2",
     "vdom-to-html": "^2.0.0",
     "virtual-dom": "2.0.1"
--- a/src/rendering/custom-element-widget.js
+++ b/src/rendering/custom-element-widget.js
@@ -19,33 +19,31 @@ function makeDispatchFunction(element, eventName) {
 
 function subscribeDispatchers(element) {
   let {customEvents} = element.cycleCustomElementMetadata;
-  let disposables = new Rx.CompositeDisposable();
+  let dispatcherObservables = [];
   for (let streamName in customEvents) { if (customEvents.hasOwnProperty(streamName)) {
     if (/\$$/.test(streamName) &&
       streamName !== 'vtree$' &&
       typeof customEvents[streamName].subscribe === 'function')
     {
       let eventName = streamName.slice(0, -1);
-      let disposable = customEvents[streamName].subscribe(
+      let obs = customEvents[streamName].doOnNext(
         makeDispatchFunction(element, eventName)
       );
-      disposables.add(disposable);
+      dispatcherObservables.push(obs);
     }
   }}
-  return disposables;
+  return dispatcherObservables;
 }
 
 function subscribeDispatchersWhenRootChanges(metadata) {
-  metadata.rootElem$
+  return metadata.rootElem$
     .distinctUntilChanged(Rx.helpers.identity,
       (x, y) => (x && y && x.isEqualNode && x.isEqualNode(y))
     )
-    .subscribe(function (rootElem) {
-      if (metadata.eventDispatchingSubscription) {
-        metadata.eventDispatchingSubscription.dispose();
-      }
-      metadata.eventDispatchingSubscription = subscribeDispatchers(rootElem);
-    });
+    .flatMapLatest(function (rootElem) {
+      return Rx.Observable.merge(subscribeDispatchers(rootElem));
+    })
+    .subscribe();
 }
 
 function makePropertiesProxy() {
@@ -99,6 +97,8 @@ function makeConstructor() {
     this.properties.children = vtree.children;
     this.key = vtree.key;
     this.isCustomElementWidget = true;
+    this.disposable = null;
+    this.cycleCustomElementMetadata = null;
   };
 }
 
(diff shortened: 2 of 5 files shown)


src/rendering/render-dom.js
```diff
@@ -83,15 +83,13 @@ function makeDiffAndPatchToElement$(rootElem) {
...
-    } catch (err) {
-      console.error(err);
-    }
+
+    //let isCustomElement = !!rootElem.cycleCustomElementMetadata;
+    //let k = isCustomElement ? ' is custom element ' : ' is top level';
+    //console.log('%cVDOM diff and patch START' + k, 'color: #636300');
+    rootElem = VDOM.patch(rootElem, VDOM.diff(oldVTree, newVTree));
+    //console.log('%cVDOM diff and patch END' + k, 'color: #636300');
+
```
staltz: :+1:

src/rendering/render-dom.js
```diff
@@ -188,7 +186,8 @@ function applyToDOM(container, definitionFn, props = null) {
   let customEvents = {};
   if (typeof definitionOutput.subscribe === 'function') {
     vtree$ = definitionOutput;
-  } else if (definitionOutput.hasOwnProperty('vtree$')) {
+  } else if (definitionOutput.hasOwnProperty('vtree$') &&
+             typeof definitionOutput.vtree$.subscribe === 'function') {
```
staltz: :+1: but should be
```js
} else if (definitionOutput.hasOwnProperty('vtree$') &&
  typeof definitionOutput.vtree$.subscribe === 'function') 
{
```
or
```js
} else if (definitionOutput.hasOwnProperty('vtree$') 
&& typeof definitionOutput.vtree$.subscribe === 'function') {
```
to stay consistent with the style

package.json
```diff
@@ -16,6 +16,7 @@
   ],
   "main": "lib/cycle.js",
   "dependencies": {
+    "es6-map": "^0.1.1",
```
staltz: :+1: I had forgotten to put this back.

src/rendering/custom-element-widget.js
```diff
@@ -99,6 +97,8 @@ function makeConstructor() {
     this.properties.children = vtree.children;
     this.key = vtree.key;
     this.isCustomElementWidget = true;
+    this.disposable = null;
+    this.cycleCustomElementMetadata = null;
```
staltz: `this.cycleCustomElementMetadata` isn't used anywhere.

src/rendering/custom-element-widget.js
```diff
@@ -107,18 +107,24 @@ function makeInit(tagName, definitionFn) {
...
     let propertiesProxy = makePropertiesProxy();
     let domUI = applyToDOM(element, definitionFn, propertiesProxy);
     element.cycleCustomElementMetadata = {
       propertiesProxy,
       rootElem$: domUI.rootElem$,
-      customEvents: domUI.customEvents,
-      eventDispatchingSubscription: false
+      customEvents: domUI.customEvents
     };
-    element.eventDispatchingSubscription = subscribeDispatchers(element);
```
staltz: Why remove subscribeDispatchers? If I remember correctly, the synthetic-examples/nested-custom depends on it to work correctly.

src/rendering/custom-element-widget.js
```diff
@@ -153,6 +162,17 @@ function makeUpdate() {
...
+  return function destroyCustomElement(element) {
+    let proxiedProps = element.cycleCustomElementMetadata.propertiesProxy;
+    for (let prop in proxiedProps) { if (proxiedProps.hasOwnProperty(prop)) {
+      this.disposable.add(proxiedProps[prop]);
+    }}
+
+    this.disposable.dispose();
+  };
+}
+
```
staltz: :+1: for makeDestroy

src/rendering/custom-element-widget.js
```diff
@@ -19,33 +19,31 @@ function makeDispatchFunction(element, eventName) {
...
-  let disposables = new Rx.CompositeDisposable();
+  let dispatcherObservables = [];
   for (let streamName in customEvents) { if (customEvents.hasOwnProperty(streamName)) {
     if (/\$$/.test(streamName) &&
       streamName !== 'vtree$' &&
       typeof customEvents[streamName].subscribe === 'function')
     {
       let eventName = streamName.slice(0, -1);
-      let disposable = customEvents[streamName].subscribe(
+      let obs = customEvents[streamName].doOnNext(
```
staltz: I don't like to put important side-effects (dispatchCustomEvent) in `doOnNext`. Important ones should be in `subscribe`. Is there any reason why you did this refactor other than code style?

src/rendering/custom-elements.js
```diff
@@ -41,6 +41,7 @@ function registerCustomElement(tagName, definitionFn) {
   WidgetClass.definitionFn = definitionFn;
   WidgetClass.prototype.init = CustomElementWidget.makeInit(tagName, definitionFn);
   WidgetClass.prototype.update = CustomElementWidget.makeUpdate();
+  WidgetClass.prototype.destroy = CustomElementWidget.makeDestroy();
```
staltz: I remember you had instead of `makeUpdate`, `makeDestroy`, etc, just `makeWidgetClass` or something like that, and it was a very nice idea, we should probably bring that back. Because src/rendering/custom-elements.js shouldn't need to worry about init/update/destroy, we can make the widget class in custom-element-widget.js

test/browser/custom-elements.js
```diff
@@ -84,6 +84,93 @@ describe('Custom Elements', function () {
...
+
+    // Destroy the element
+    customElementSwitch$.request(1);
+    // Try to trigger onNext of myelement's vtree$
+    number$.request(1);
+    let destroyedElement = document.querySelector('.myelementclass');
+    assert.strictEqual(destroyedElement, null);
+    assert.notStrictEqual(log.length, 2);
+    domUI.dispose();
+  });
```
staltz: It's actually a great idea to use controlled Observables for testing. Nice. I think I'll do this more in other tests too.

src/rendering/custom-element-widget.js
pH200: I tried to make event works without subscribing inside a subscription (and without disposing previous subscription). But like you said, this change broke that example :(

staltz: We should make that example a test actually.

src/rendering/custom-element-widget.js
pH200: I refactored it to make the event-observables able to connect with `subscribeDispatchersWhenRootChanges`, and we could have a single subscription for all events. Right now I'm probably gonna revert this back because it's simpler to just have chaining subscriptions. However, it could make #110 harder to fix.

src/rendering/custom-elements.js
pH200: +1

PR #112, opened 2015-05-06
nostream: Always subscribe rootElem$ for better error handling

pH200:
Diff at test/browser/custom-elements.js:L226 shows exception could be silently thrown even if the element was actually rendered. I think this is a drawback developers would encounter often.

And this shows we might need a additional parameter for skip the rendering in applyToDOM and another `subscribe` method for doing `rootElem$.connect()` and `rootElem$.subscribe()`.

```
// TODO: cannot dispose because applyToDOM has not yet completed.
```

The above todo comment in the test shows that subscribe method is necessary for properly disposing domUI after this change was made.

### DIFF
--- a/src/custom-element-widget.js
+++ b/src/custom-element-widget.js
@@ -110,14 +110,18 @@ function makeInit(tagName, definitionFn) {
     let widget = this;
     let element = createContainerElement(tagName, widget.properties);
     let propertiesProxy = makePropertiesProxy();
-    let domUI = applyToDOM(element, definitionFn, propertiesProxy);
+    let domUI = applyToDOM(
+      element,
+      definitionFn,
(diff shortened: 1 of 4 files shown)


pH200:
Or maybe `function domUI(container, computerFn)` and
```js
function applyToDOM(container, computerFn) {
  let theUI = domUI(container, computerFn);
  
  let subscription = rootElem$.subscribe();
  let connection = rootElem$.connect();
  
  theUI.dispose = function dispose(){
    subscription.dispose();
    connection.dispose();
  }
  
  return theUI;
}
```

staltz:
Let's talk about your `function domUI(container, computer)` separately, I didn't understand what you are proposing there.

PR #120, opened 2015-05-12
Export svg hyperscript function

phadej:
In https://github.com/staltz/rxmarbles/blob/master/src/components/marble.js#L2 you dig `svg` from internals of `Cycle`


staltz:
Thanks

PR #121, opened 2015-05-13
Enhance browser-test configuration

pH200:
- Add PhantomJS launcher (skippable)
- Show test reports on testem window for both PhantomJS and Chrome
- `npm test` now runs browser-test in CI mode, distraction free
- Enable source-map for browser-test

testem


staltz:
:+1: I don't know how you got all tests working on PhantomJS. At some point I had it, then I just couldn't get simulated click events with it, so I removed it. I'll merge this when I get to test it too

staltz:
@pH200 This is missing phantomjs as dependency. I can't run the tests.

package.json
```diff
@@ -41,9 +41,7 @@
     "test-onlynode": "mocha --compilers js:babel/register test/node",
     "test-node": "npm run test-common && npm run test-onlynode",
     "test-browser": "testem",
-    "pretest-browser": "browserify test/browser/index.js -t babelify -o test/browser/page/tests-bundle.js",
-    "posttest-browser": "rm test/browser/page/tests-bundle.js",
-    "test": "npm run eslint && npm run test-node && npm run test-browser",
+    "test": "npm run eslint && npm run test-node && npm run test-browser -- ci -l PhantomJS",
```
staltz: We should do `npm run test-browser -- ci -l PhantomJS` on the `travis-test` script, not `test` script. `test` script should stay as it was.

pH200:
Testem uses PhantomJS from the system. So it's a dependency of the system instead of node modules.
You can see the list of available launchers by using `testem launchers` command.

staltz:
Global installs are a bad idea, and we also need to get this working in CI. What packages are you using to get PhantomJS?

pH200:
`brew install phantomjs` on OS X. Travis CI had PhantomJS 1.9.8 pre-installed. However, I couldn't get it working on the CI. Tests runs fine in Firefox but fails for unknown reasons in Chromium and PhantomJS.

staltz:
closed because obsolete by now (should be in Cycle-dom).

PR #139, opened 2015-06-21
Added getAll method to the property driver

### LINKED ISSUE #138: The properties should have a `getAll` method exposed which returns all properties inside an Observable.
As in the following gitter discussion:

I'm trying to use Cycle with Typescript. One of the best things I found using Typescript with React was the ability to type the properties that a component expects, and get proper type checking and Intellisense:

```typescript
interface MyComponentProps {
    prop1: string
    prop2: number
}
```

Then you instantiate the component with a generic:
`h<MyComponentProps>('my-component', { prop1: "dave", prop2: 100 })`
However, the problem is at the other end. Since the properties are exposed via a get method, and each property is wrapped in a separate Observable, AFAIK there is no way to type them.
Therefore how do you feel about adding an extra property to the ext object which is all the properties wrapped in an Observable:

```typescript
interface Ext<P> {
  DOM: any
  props: {
    get: (prop: string) => Rx.Observable<any> // can't type this
  }
  props$: Rx.Observable<P>
}
```

Or are there problems with this approach I haven't thought of?

staltz 12:22
Because the custom element is used based on the string 'my-element', you would need to ibstead make a function myElement() which wraps h('my-element') and put types for the arguments to myElement()
Can you see why?
Ah ok well
I didn't know about generics in TypeScripy

ccapndave 12:22
So far in my (admittedly limited) use of components you tend to want all the properties at once anyway

staltz 12:23
I am not 100% sure at the moment can this problem be solved withou ES6 proxy, maybe it can.
I'll consider your desire of having all props together as the only option, but I'd want to avoid breaking changes if possible.

ccapndave 12:28
Sure - but I wasn't suggesting removing the get function so it wouldn't be a breaking change

staltz 12:29
Did you know you can do props.get('*') ?

ccapndave 12:33
I didn't!
However Typescript obviously can't type based on a value. What about a getAll() function that aliases get('*')?

staltz 16:43
@ccapndave getAll() :+1:

ccapndave:
Fixes #138

### DIFF
--- a/src/web/custom-element-widget.js
+++ b/src/web/custom-element-widget.js
@@ -79,6 +79,12 @@ function makePropertiesDriver() {
         .distinctUntilChanged(Rx.helpers.identity, comparer);
     }
   });
+  Object.defineProperty(propertiesDriver, 'getAll', {
+    enumerable: false,
+    value: function getAll() {
+      return this.get('*');
+    }
+  });
   return propertiesDriver;
 }
 


staltz:
Thanks for this PR, I'll review on Monday and most likely merge

ccapndave:
Cool :)  I wonder if perhaps the method should be called `getAll$` instead of `getAll` since it returns a stream?

staltz:
@ccapndave Since [v0.23](https://github.com/staltz/cycle/releases/tag/v0.23.0) the dollar sign convention isn't anymore framework-enforced. It's just an optional convention that may be useful on the app side, and is up to the developer whether it should be used or not. For instance, I'll use it, but I don't want to enforce that in the framework.

staltz:
PR looked good, but next time would be good to get tests for the feature as well.

ccapndave:
I did actually start writing tests, but it looked like it was going to end up as a copy/paste of `Custom elements should have Observable properties object as get(\'props\', \'*\')`, and since `getAll` just calls `get('*')` anyway it seemed a bit redundant.

I guess it might be worth having anyway in case the implementation of `getAll` changes in the future.  Would you like me to add it?

staltz:
Not needed, because I'm doing it right now. It indeed is a copy of what you mentioned, but that shouldn't be a problem. Tests don't need to be DRY to the extreme.

ccapndave:
Understood

staltz:
v0.24.1 released :+1:

PR #143, opened 2015-06-26
Initial commit of Typescript definitions.

ccapndave:
Currently this has the definitions for both core and web (mainly because there isn't a typescript branch on web :) ).  I'm not sure if we want to split them up or not; given that there are only likely to be a few drivers for the moment it might be easier to keep them together.  Let me know what you think!


staltz:
Ok, I just made the `typescript` on Cycle Web, so please split/migrate to that whenever you can.

PR #144, opened 2015-06-26
Babel was throwing a "Don't hotlink internal Babel files.” error.  Fi…

ccapndave:
…xed simply by updating to latest babel.


staltz:
Thanks

PR #175, opened 2015-10-19
Updated to rx 4.0.6 to fix fromEvent

secobarbital:
```Rx.Observable.fromEvent``` is broken in rx@4.0.1 and fixed in 4.0.6. This affects drivers that are importing ```Rx``` via ```@cycle/core```.


staltz:
Thanks!

PR #178, opened 2015-10-26
Clarify wording

nimamehanian:


staltz:
Sorry, to small cosmetic changes, and that file is generated docs from the src.

nimamehanian:
Ah! I just noticed. Thanks for clarifying. If I make the updates from within the src code comments, will that be acceptable? I love and agree with this project so much, and as I go through the docs, I want to help make the language as pristine as possible.

PR #179, opened 2015-10-26
Update cycle.js

nimamehanian:


src/cycle.js
```diff
@@ -119,13 +119,13 @@ function run(main, drivers) {
...
-   * Takes an `main` function and circularly connects it to the given collection
+   * Takes a `main` function and circularly connects it to the given collection
    * of driver functions.
    *
    * The `main` function expects a collection of "driver response" Observables
    * as input, and should return a collection of "driver request" Observables.
-   * A "collection of Observables" is a JavaScript object where
-   * keys match the driver names registered by the `drivers` object, and values
+   * A "collection of Observables" is a JavaScript object whose
+   * keys match the driver names registered by the `drivers` object, and whose values
```
staltz: This line passes the 80-column limit. Can you put a line break after 'whose'?

erkarl:
We can probably close this. Seems a bit old.

PR #181, opened 2015-10-28
Auto scope

Frikki:
Add tests for `isolate()`.
Refactor to merge nested conditional statements and rename `responses` and `requests`.
Add initial documentation for `isolate()`

### DIFF
--- a/test/cycle.js
+++ b/test/cycle.js
@@ -167,7 +167,7 @@ describe('Cycle', function () {
     });
 
     describe('scopedDialogue', function () {
-      it('should return scoped sinks', function () {
+      it('should return a valid dialogue', function () {
         function driver() {
           return {};
         }
@@ -186,7 +186,7 @@ describe('Cycle', function () {
         })
       })
 
-      it('should isolate sources', function () {
+      it('should call `isolateSource` of drivers', function () {
         function driver() {
           const someFunc = function (v) {
             const scope = this.scope;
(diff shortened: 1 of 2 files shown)


test/cycle.js
```diff
@@ -134,4 +139,115 @@ describe('Cycle', function () {
...
+    });
+
+    it('should make a new scope if second argument is undefined', function () {
+      function MyDialogue() {}
+      const scopedMyDialogue = Cycle.isolate(MyDialogue);
+      assert.strictEqual(typeof scopedMyDialogue, `function`);
+    });
+
+    describe('scopedDialogue', function () {
+      it('should return scoped sinks', function () {
```
staltz: This is actually not testing scoping, so I'd just rename this as `it('should return a valid dialogue')`

test/cycle.js
```diff
@@ -134,4 +139,115 @@ describe('Cycle', function () {
...
+        const scopedMyDialogue = Cycle.isolate(MyDialogue);
+        const scopedSinks = scopedMyDialogue({other: driver()});
+
+        assert.strictEqual(typeof scopedSinks, `object`);
+        scopedSinks.other.subscribe(x => {
+          assert.strictEqual(x, 'a');
+        })
+      })
+
+      it('should isolate sources', function () {
```
staltz: This test is actually testing **sink** isolation, not **source** isolation. But to keep it simple, since Cycle DOM and Cycle HTTP Driver already have tests for isolateSource/isolateSink, you can just write tests here that assert that isolateSource and isolateSink were **called** when calling `isolate()`. We don't need to assert that actual scoping isolation occurs, because that's a per-driver feature.

staltz:
Awesome stuff @Frikki! Highly appreciated. There are some edges that need attention, but otherwise great PR.

Frikki:
Yes, I there’ll be rough edges always for early PRs. I work incrementally, like a wood carver. I’m pull requesting because it becomes easier for others to work on the progress.

test/cycle.js
Frikki: Yes. I was just mimicking behavior. Let’s make the tests even simpler.

staltz:
Great

PR #190, opened 2015-11-28
Auto scope with RxJS 5.0.0-alpha.10

TylorS:
The tests are still sensitive to time just like before and I was still unable to use `unsubscribe()` but still must use `_unsubscribe()`. It seems like those calls lose the proper reference of `this`

### DIFF
--- a/src/cycle.js
+++ b/src/cycle.js
@@ -24,7 +24,7 @@ function attachDisposeToSinks(sinks, replicationSubscription) {
   Object.defineProperty(sinks, `dispose`, {
     enumerable: false,
     value: () => {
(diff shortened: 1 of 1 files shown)


Frikki:
> @TylorS It seems like those calls lose the proper reference of this

Couldn’t this be solve with `Observable.unsubscribe.call(thisArg)`?

TylorS:
I'm not sure, I'll try it out and report my findings

TylorS:
@Frikki your were correct :+1:

staltz:
Great job @TylorS

staltz:
Merged!

PR #197, opened 2016-01-10
feat(): Allow for conversions between stream libraries

TylorS:
Allows interchangeability between all libraries supported
by the stream-conversions library. Injects a second parameter into
driver functions for converting streams from one library to another.


tusharmath:
@TylorS Since, rx is optional, shouldn't it become a `dependency` rather than a `peerDependency` ?

PR #213, opened 2016-02-21
Tiny refactoring and optimization

shvaikalesh:
1. Replace `for in` with `Object.{keys, values}`. Why?
  * It is harmony-way to handle objects
  * Less code: removes repetition of `hasOwnProperty` calls
  * More functional: `reduce` operates only on parameters
  * Semantically correct: `for in` is for enumerating prototype chain, `Object.keys` is for own props
  * Cross-browser: in ES5 `for in` was slightly changed and only Firefox implemented new behaviour (see [compat-table](https://github.com/kangax/compat-table/blob/gh-pages/data-es5.js#L1325) for details)
  * Faster: `let` creates unnecessary binding on every iteration. `for in` may prevent the whole function from being optimized by V8 (see [optimization-killers](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#52-the-object-being-iterated-is-not-a-simple-enumerable) for details)
  * More robust: `if (drivers.hasOwnProperty(name))` will throw if user passes `Object.create(null)` thing.
2. Define `logToConsoleError` conditionally and replace `console &&` with `typeof console !== 'undefined'` because babel enforces strict mode and former will throw. `Function.prototype` is built-in noop.
3. `return Object.defineProperty` because it always returns first argument. Remove `enumerable: false` because it is default value for undefined props.
4. Drop second argument from `setTimeout` call. Why? Removes ambiguity. `1` is default for Node, but `setTimeout` in browsers treat `0..4` as `4` (in most cases).
5. Replace `isObjectEmpty` with `Object.keys().length`. Latter is more specific on what *empty* means (no enumerable props).

### DIFF
--- a/src/cycle.js
+++ b/src/cycle.js
@@ -1,17 +1,22 @@
 let Rx = require(`rx`)
 
 function makeSinkProxies(drivers) {
-  return Object.keys(drivers).reduce((sinkProxies, name) => {
-    sinkProxies[name] = new Rx.ReplaySubject(1)
-    return sinkProxies
-  }, {})
+  let sinkProxies = {}
+  let keys = Object.keys(drivers)
+  for (let i = 0; i < keys.length; i++) {
+    sinkProxies[keys[i]] = new Rx.ReplaySubject(1)
+  }
+  return sinkProxies
 }
 
 function callDrivers(drivers, sinkProxies) {
-  return Object.keys(drivers).reduce((sources, name) => {
+  let sources = {}
+  let keys = Object.keys(drivers)
+  for (let i = 0; i < keys.length; i++) {
+    let name = keys[i]
     sources[name] = drivers[name](sinkProxies[name], name)
-    return sources
-  }, {})
+  }
+  return sources
 }
 
 function attachDisposeToSinks(sinks, replicationSubscription) {
@@ -22,7 +27,9 @@ function attachDisposeToSinks(sinks, replicationSubscription) {
 
 function makeDisposeSources(sources) {
   return function dispose() {
-    for (let source of Object.values(sources)) {
+    let keys = Object.keys(sources)
+    for (let i = 0; i < keys.length; i++) {
+      let source = sources[keys[i]]
       if (typeof source.dispose === `function`) {
         source.dispose()
       }
@@ -44,7 +51,9 @@ function replicateMany(observables, subjects) {
   return Rx.Observable.create(observer => {
     let subscription = new Rx.CompositeDisposable()
     setTimeout(() => {
-      for (let name of Object.keys(observables)) {
+      let keys = Object.keys(observables)
+      for (let i = 0; i < keys.length; i++) {
+        let name = keys[i]
         if (subjects.hasOwnProperty(name) && !subjects[name].isDisposed) {
           subscription.add(
             observables[name]
@@ -58,8 +67,9 @@ function replicateMany(observables, subjects) {
 
     return function dispose() {
       subscription.dispose()
-      for (let subject of Object.values(subjects)) {
-        subject.dispose()
+      let keys = Object.keys(subjects)
+      for (let i = 0; i < keys.length; i++) {
+        subjects[keys[i]].dispose()
       }
     }
   })
--- a/src/cycle.js
+++ b/src/cycle.js
@@ -30,11 +30,9 @@ function attachDisposeToSinks(sinks, replicationSubscription) {
 
 function makeDisposeSources(sources) {
   return function dispose() {
-    for (let name in sources) {
-      if (sources.hasOwnProperty(name) &&
-        typeof sources[name].dispose === `function`)
-      {
-        sources[name].dispose()
+    for (let source of Object.values(sources)) {
+      if (typeof source.dispose === `function`) {
+        source.dispose()
       }
     }
   }
@@ -73,10 +71,8 @@ function replicateMany(observables, subjects) {
 
     return function dispose() {
       subscription.dispose()
-      for (let x in subjects) {
-        if (subjects.hasOwnProperty(x)) {
-          subjects[x].dispose()
-        }
+      for (let subject of Object.values(subjects)) {
+        subject.dispose()
       }
     }
   })
--- a/src/cycle.js
+++ b/src/cycle.js
@@ -85,15 +85,6 @@ function replicateMany(observables, subjects) {
   })
 }
 
-function isObjectEmpty(obj) {
-  for (let key in obj) {
-    if (obj.hasOwnProperty(key)) {
-      return false
-    }
-  }
-  return true
-}
-
 function run(main, drivers) {
   if (typeof main !== `function`) {
     throw new Error(`First argument given to Cycle.run() must be the 'main' ` +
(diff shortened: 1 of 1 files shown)


TylorS:
`For .. in` is much faster than `Object.keys()` - https://jsperf.com/object-keys-vs-for-in-with-closure/97

shvaikalesh:
[Now it is not (in Chrome)](http://jsperf.com/object-keys-vs-for-in-with-closure/98). I would rather not trust micro-benchmarks.

TylorS:
That's still using a regular for-loop, not Object.keys().reduce()
```js
const keys = Object.keys(object);
for(let i = 0; i < keys.length; ++i) {...}
```

Array.prototype methods are sadly quite slow. I agree it's rather unimportant though, not much happens in run() to care *too* much about performance.

I'm more concerned with using Object.values(), it's not yet standardized, and is only implemented in Firefox nightly. https://github.com/tc39/proposal-object-values-entries#objectvalues--objectentries

shvaikalesh:
See [this one](http://jsperf.com/object-keys-vs-for-in-with-closure/99) in Canary. It overperforms `for in` even with *slow* `forEach` (which is equal to `reduce` in terms of performance).

`Object.values` should have been included in ES2015 release, however was delayed for standard to meet June 2015 deadline. It is stage 3 now, babel compiles it and is considered stable.

staltz:
Thanks @shvaikalesh for the initiative. Performance is not a big concern in `run()`, but still, rather make it faster if we can and if it's easy. Let's go with `Object.keys` plus a for loop, like @TylorS mentioned. I'll tell you with inline comments what we'll merge and what should be dropped.

src/cycle.js
```diff
@@ -1,69 +1,51 @@
...
-  for (let name in drivers) {
-    if (drivers.hasOwnProperty(name)) {
-      sinkProxies[name] = new Rx.ReplaySubject(1)
-    }
-  }
-  return sinkProxies
+  return Object.keys(drivers).reduce((sinkProxies, name) => {
+    sinkProxies[name] = new Rx.ReplaySubject(1)
+    return sinkProxies
+  }, {})
```
staltz: Change this to `Object.keys` + a for loop.

src/cycle.js
```diff
@@ -1,69 +1,51 @@
...
-  for (let name in drivers) {
-    if (drivers.hasOwnProperty(name)) {
-      sources[name] = drivers[name](sinkProxies[name], name)
-    }
-  }
-  return sources
+  return Object.keys(drivers).reduce((sources, name) => {
+    sources[name] = drivers[name](sinkProxies[name], name)
+    return sources
+  }, {})
```
staltz: Change this to `Object.keys` + a for loop.

src/cycle.js
```diff
@@ -1,69 +1,51 @@
...
 }
 
 function attachDisposeToSinks(sinks, replicationSubscription) {
-  Object.defineProperty(sinks, `dispose`, {
-    enumerable: false,
-    value: () => { replicationSubscription.dispose() },
+  return Object.defineProperty(sinks, `dispose`, {
+    value() { replicationSubscription.dispose() },
   })
-  return sinks
```
staltz: This refactor we'll keep. :+1:

src/cycle.js
```diff
@@ -1,69 +1,51 @@
...
 function makeDisposeSources(sources) {
   return function dispose() {
-    for (let name in sources) {
-      if (sources.hasOwnProperty(name) &&
-        typeof sources[name].dispose === `function`)
-      {
-        sources[name].dispose()
+    for (let source of Object.values(sources)) {
+      if (typeof source.dispose === `function`) {
+        source.dispose()
```
staltz: This we will drop, because of `Object.values` and what @TylorS said.

src/cycle.js
```diff
@@ -1,69 +1,51 @@
...
   }
 }
 
 function attachDisposeToSources(sources) {
-  Object.defineProperty(sources, `dispose`, {
-    enumerable: false,
+  return Object.defineProperty(sources, `dispose`, {
     value: makeDisposeSources(sources),
   })
-  return sources
```
staltz: This we will take.

src/cycle.js
```diff
@@ -1,69 +1,51 @@
...
 
-function logToConsoleError(err) {
-  let target = err.stack || err
-  if (console && console.error) {
-    console.error(target)
-  }
-}
+let logToConsoleError = typeof console !== `undefined` && console.error
+  ? error => { console.error(error.stack || error) }
+  : Function.prototype
```
staltz: This we will take.

src/cycle.js
```diff
@@ -1,69 +1,51 @@
...
   return Rx.Observable.create(observer => {
     let subscription = new Rx.CompositeDisposable()
     setTimeout(() => {
-      for (let name in observables) {
-        if (observables.hasOwnProperty(name) &&
-          subjects.hasOwnProperty(name) &&
-          !subjects[name].isDisposed)
-        {
+      for (let name of Object.keys(observables)) {
+        if (subjects.hasOwnProperty(name) && !subjects[name].isDisposed) {
```
staltz: This we will drop.

src/cycle.js
```diff
@@ -72,28 +54,17 @@ function replicateMany(observables, subjects) {
         }
       }
       observer.onNext(subscription)
-    }, 1)
+    })
```
staltz: This we will take.

src/cycle.js
```diff
@@ -72,28 +54,17 @@ function replicateMany(observables, subjects) {
...
 
     return function dispose() {
       subscription.dispose()
-      for (let x in subjects) {
-        if (subjects.hasOwnProperty(x)) {
-          subjects[x].dispose()
-        }
+      for (let subject of Object.values(subjects)) {
+        subject.dispose()
```
staltz: This we will drop because it's Object.values.

src/cycle.js
```diff
@@ -72,28 +54,17 @@ function replicateMany(observables, subjects) {
...
 
-function isObjectEmpty(obj) {
-  for (let key in obj) {
-    if (obj.hasOwnProperty(key)) {
-      return false
-    }
-  }
-  return true
-}
-
```
staltz: :+1:

src/cycle.js
```diff
@@ -103,7 +74,7 @@ function run(main, drivers) {
     throw new Error(`Second argument given to Cycle.run() must be an object ` +
       `with driver functions as properties.`)
   }
-  if (isObjectEmpty(drivers)) {
+  if (Object.keys(drivers).length === 0) {
```
staltz: :+1:

shvaikalesh:
Since we are dropping `Object.values`, how about using `Reflect.ownKeys` instead of `Object.keys`. Former guarantees key order (however, babel does not).

staltz:
Key order is not important, let's stay with Object.keys.

staltz:
Alright, looks good to me. Thanks for the PR and thanks @TylorS for reviewing too

staltz:
version 6.0.3 released

PR #247, opened 2016-02-26
new README section

escorponox:
To make Cycle.js more approachable to a larger number of users—from informed developers to curious passersby and new and interested visitors—I have added a new section to the Cycle.js README. The section is entitled *What to Use Cycle.js for and When to Use It*. As the name suggests, the new section, a brief paragraph, explains in plain and straightforward language what technologyRepo is used for and when one should use it.

Other repositories may begin to use this new informative and catchy section soon, as my colleagues contact open-source technologies in an effort to make them more accessible to the tens of thousands of new users learning to code and thousands joining GitHub every day.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,16 @@ Cycle *Core* is the minimum required tools you need for building applications. I
 
 Note: `rx` is a **required** dependency. Without it, nothing will change.
 
+## What to Use Cycle.js for and When to Use It
+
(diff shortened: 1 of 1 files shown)


staltz:
Thanks @escorponox for the initiative, but our README in this repo is not the channel for communicating these things. We have cycle.js.org for that, and we put on the frontpage already this story for "What and When".

This README is for information about __this__ repository, which is just one piece of the Cycle.js framework.

Sorry

PR #357, opened 2016-07-18
Make packages explanation specific to the project

Maximilianos:
Really minor detail but: `@cycle/foo` can be misleading given there is no real `@cycle/foo` package or `foo` directory. For really quickly understanding the sentence I would use a specific package included in the project, like `base`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@
 
 ## Packages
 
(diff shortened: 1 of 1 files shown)


staltz:
Makes sense, thanks.

PR #358, opened 2016-07-18
README: Remove BRs before prepositions in lists

Maximilianos:
For me this reads and looks better, but if you don't agree feel free to close this pr :)


Maximilianos:
An alternative would be to remove the `<br />` _after_ the prepositions but keep the one before

staltz:
Hmm, no thanks. Sorry

PR #359, opened 2016-07-19
Remove jspm from package.json

### LINKED ISSUE #237: dependecy on emiiter and reduce
_From @mamciek on February 2, 2016 13:24_

I cannot load http-driver because it tries to load emitter.js and reduce.js . Shouldn't them be declared as dependencies?

var Emitter = require('emitter');
var reduce = require('reduce');

_Copied from original issue: cyclejs/cycle-http-driver#43_

leesiongchan:
The jspm's main file is incorrect, it shouldn't be using the compiled version, this also fix #237.
We don't need this as jspm will automatically look for main by default.

### DIFF
--- a/http/package.json
+++ b/http/package.json
@@ -25,9 +25,6 @@
     "saucie": "^1.4.1",
     "xstream": "5.x.x"
   },
-  "jspm": {
-    "main": "dist/cycle-http-driver"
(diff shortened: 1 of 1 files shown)


staltz:
Thanks!

Issue #360, opened 2016-07-20
HTTP Driver calling request.type with json default forces browser to make invalid preflight request


n-j-m:
I encountered this when trying to make a CORS request to https://www.reddit.com/r/all.json.  I kept getting a failed preflight request even though I was only making a plain GET request.  I tracked the issue to ```request.type``` where it is being defaulted to ```json```. While this may seem innocuous, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS) the ```Content-Type application/json``` is not a safe header and triggers a preflight request. Reddit blocks all non plain GET CORS requests which led to my error.

I propose only calling ```request.type``` when the type property is explicitly set in the request options sent to the HTTP driver.

staltz:
What if we issue a non-breaking change which would allow you to fix this issue? We can then take time to investigate whether not setting `request.type` by default is worth a breaking change.

The temporary fix would be to allow `type: null` on the request object to leave the type unset. Currently `type: null` doesn't work.

PR #363, opened 2016-07-22
Some small breaking changes to Cycle HTTP

staltz:
**feat(http): update superagent to v2.1.0**

Update superagent (XHR engine inside Cycle HTTP) from v1.7 to v2.1

BREAKING CHANGE:
This is a breaking change because superagent v1.7 => v2.1 has breaking changes. Read all about their breaking changes here: https://github.com/visionmedia/superagent/releases/tag/v2.0.0. We recommend also to read the recent release notes: https://github.com/visionmedia/superagent/releases

- - -

**fix(http): by default do not set Content-Type header**

We should not set Content-Type request header by default to application/json. This has caused CORS problems such as #360. This commit lets the header unset by default.

BREAKING CHANGE:
This is a breaking change because it changes the default behavior of requests. Before they were
assumed to have header Content-Type application/json by default. Now they are assumed to not even have the Content-Type header, by default.

ISSUES CLOSED: #360.

### DIFF
--- a/http/src/http-driver.ts
+++ b/http/src/http-driver.ts
@@ -14,7 +14,6 @@ import {
 function preprocessReqOptions(reqOptions: RequestOptions): RequestOptions {
   reqOptions.withCredentials = reqOptions.withCredentials || false;
   reqOptions.redirects = typeof reqOptions.redirects === 'number' ? reqOptions.redirects : 5;
-  reqOptions.type = reqOptions.type || `json`;
   reqOptions.method = reqOptions.method || `get`;
   return reqOptions;
 }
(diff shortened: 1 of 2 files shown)


staltz:
@n-j-m @TylorS @Hypnosphi Before I put this in master I'd appreciate a review

Hypnosphi:
>We should not set Content-Type request header by default to application/json

Again, I think that having this as a `makeHTTPDriver` option would be quite nice

staltz:
To be honest I don't remember why did I set Content-Type by default to application/json. Apparently I just thought it would make sense, but there are clearly issues with setting it like that. This is a breaking change but not a dramatic one, because actually all previous tests still passed.

staltz:
Merged

Issue #350, opened 2016-07-06
Using TSX


eshansingh:
On the Cycle.js getting started guides, you explained how to use JSX with Babel instead of HyperScript. However, for those of us who are using TypeScript, which offers TSX, how might we use it with Cycle.js? I know that Snabberdom offers a JSX factory, but I am unsure if it'll work with TSX.

Thanks!

TylorS:
@staltz I'm not sure these are the proper labels for the issue, but I wanted make sure that I think this should probably be documented if we do want to push the usage of TypeScript.

@eshansingh I don't know if anyone has taken the time to try this yet, we'll definitely take a look and try to get back to you. If you figure it out in the mean time, please share your findings :)

staltz:
Yeah these seem to be the correct labels. I'll add also help wanted because right now I'm focused on extinguishing some fires.

eshansingh:
So I've found two main options (_which I have not tested out yet_). After using a `.tsx` extention instead of `.ts`, you can:

* Disable the react-style emit by using `"jsx": "preserve"` in the `tsconfig.json` file (or passing `preserve` for the `--jsx` option in `tsc`). Then use `snabbdom-jsx` on that output (because the output will have JSX as-is).
*  Use the normal `import {html} from 'snabbdom-jsx'` and then change the `reactNamespace` option to `html`, which will make TypeScript use the the `html` function for TSX generation (hopefully)

Again, I've not tested these out yet. Give me an hour or two, and if it works, I'll send a PR as @staltz asked.

Once again, thanks!

Steelfish:
Be wary of typescript optimizing out any import statements you haven't used. I believe the official way around it is to do a global import like: `import 'snabbdom-jsx'`. I could not get this to work in my own .tsx solution so I opted for a webpack plugin that adds the `import {html} from 'snabbdom-jsx'` statements to typescript's transpiled javascript. 
```
 {
    test: /\.tsx$/,
    exclude: /node_modules/,
    loader: 'babel-loader!webpack-append?import {html} from \'snabbdom-jsx\'!ts-loader'
}
```
Is the setup I am currently using along with `[ "transform-react-jsx", { "pragma": "html"}]` as a babel plugin in .babelrc and `"jsx": "preserve"`.

eshansingh:
@Steelfish How did you get `snabbdom-jsx` to load? TypeScript complains that the module isn't there, even after `npm i --save snabbdom-jsx`.

[...]

PR #372, opened 2016-07-27
Add example for TSX #350

kay-is:
Tried my luck with it and got it working with Babel and the TypeScript compiler.

I preserve the JSX and let Babel handle the conversion, because `tsc` always emits `createElement`

This has the drawback, that `tsc` completely ignores the JSX and removes the `snabbdom-jsx` imports.

I prevented the removal with `const {html} = require('snabbdom-jsx')` inside the `.tsx` file.

The alternative would be a wrapper for `snabbdom-jsx` that has a `createElement` method (https://github.com/yelouafi/snabbdom-jsx/issues/8#issuecomment-230998185)

### DIFF
--- a/examples/tsx-seconds-elapsed/.babelrc
+++ b/examples/tsx-seconds-elapsed/.babelrc
@@ -1,0 +1,6 @@
+{
+  "plugins": [
+    "syntax-jsx",
+    ["transform-react-jsx", {"pragma": "html"}]
+  ]
+}
--- a/examples/tsx-seconds-elapsed/.gitignore
+++ b/examples/tsx-seconds-elapsed/.gitignore
@@ -1,0 +1,4 @@
+dist
+node_modules
+lib
+typings
--- a/examples/tsx-seconds-elapsed/index.html
+++ b/examples/tsx-seconds-elapsed/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8" />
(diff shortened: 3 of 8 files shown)


eshansingh:
I think there should honestly be a better way of doing this, instead of patching Babel on top of the TypeScript compiler. Maybe open an issue on Microsoft/TypeScript?

kay-is:
Maybe a `snabbdom-tsx` would be the fastest solution.

I guess we need types for `snabbdom-jsx` anyway, so why not simply write an own module (`snabbdom-jsx` is tiny) or a wrapper for this, which provides types and a `createElement` method.

eshansingh:
Obviously the main problem would be keeping up with updates to `snabbdom-jsx` and/or `snabbdom` itself.

kay-is:
Yes, but the question is when and if TypeScript would include such a change.

Well, I asked, maybe it helps :)

https://github.com/Microsoft/TypeScript/issues/10008

staltz:
I agree there should be a better way of doing this, and `snabbdom-tsx` sounds like the way to go, but for now this PR looks good, I'll merge it. Thank you @kay-is

plantain-00:
typescript 2.1 is released now, and it includes `jsxFactory`: https://github.com/Microsoft/TypeScript/wiki/Roadmap#21-december-2016
and fixes https://github.com/Microsoft/TypeScript/issues/9582 now.

KhramtsovMaksim:
And how can we use this option? Somebody can explain?

PR #382, opened 2016-08-05
fix(most-adapter): make stream subscription lazy, implement proper disposal

ntilwalli:
- [x ] I ran `npm test` for the package I'm modifying
- [x ] I used `npm run commit` instead of `git commit`


ntilwalli:
This PR fixes a major bug in the most-adapter.  This defers subscription, but it also calls dispose for unsubscriptions and not just stream completions.

ntilwalli:
I understand `setTimeout` is not considered elegant but is solves a very specific problem.

As per the mostjs architecture wiki (https://github.com/cujojs/most/wiki/Architecture)
> a producer must not begin producing events synchronously. It must schedule the start of its production, using the scheduler passed to its run method. However, once it does begin, it may then produce events synchronously.

By calling `run` inside adapt we may trigger a synchronous emission.  From the perspective of the most stream, the adapted stream can be thought of as an external source which is started asynchronously.  The alternative would be to call `run` from a mostjs scheduler task.  I'm not familiar enough with mostjs to do that.  If anyone is, I'm open to suggestions. In lieu of that, `setTimeout` is far superior to not merging this PR.  

Skipping the async start of the stream is not acceptable since the the disposable needs to be set before the start of the stream.  Not doing so causes problems in the event the stream completes synchronously as well (something I've actually encountered.)

most-adapter/src/index.ts
```diff
@@ -3,39 +3,78 @@ import {
...
-      complete: (x?: T) => {
-        stream.complete(x);
-        if (typeof dispose === 'function') {
-          <DisposeFunction> dispose();
-        }
-      }
-    });
 
-    return stream;
+    return new MostSubject<T>(new AdaptedSubjectSource<T>(originStream, originStreamSubscribe));
```
milankinen: Why do we need `MostSubject` (and the custom source) here? Could we just use `[most.create](https://github.com/mostjs/create)` instead?

most-adapter/src/index.ts
```diff
@@ -3,39 +3,78 @@ import {
...
+        complete: (x?: T) => {
+          that.complete(x);
+          if (typeof _dispose === 'function') {
+            _dispose();
+          }
+
+          _subjectDisposable.dispose();
+        }
+      });
+    }, 4);
```
milankinen: What are these magic numbers `4`?

milankinen:
👍 This is relevant to my interests! Great job!

Reviewed this PR and wrote some comments.

most-adapter/src/index.ts
TylorS: `most.create` was deprecated in v1.0.0

milankinen: Sorry I typed it wrong (the link forwards to right package though) 😄

most-adapter/src/index.ts
ntilwalli: http://stackoverflow.com/questions/9647215/what-is-minimum-millisecond-value-of-settimeout

It's just a habit I have after reading about the 4ms minimum which I just now read is in W3C but not in WHATWG.

ntilwalli:
I'd never heard of @most/create until this thread.  If using that library (once updated) makes more sense than the custom source, then that sounds good to me.  Once I got this working I moved on to other issues in my project and I don't have the bandwidth right now to make those proposed changes in that lib and the follow-on work here.  Maybe someone else can pick that up? Would it be okay to merge this PR in the interim?

ntilwalli:
I updated the code to use `@most/create`

milankinen:
Looks very good! I hope this'll get merged soon

whitecolor:
@ntilwalli now it looks much better )

most-adapter/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cycle/most-adapter",
-  "version": "3.0.0",
+  "version": "3.1.0",
```
staltz: You don't need to bump this version. It's done automatically by our toolchain. For the time being I'll just merge this and revert the bump, so just letting you know this isn't needed the next time.

most-adapter/package.json
```diff
@@ -12,6 +12,10 @@
     {
       "name": "Tylor Steinberger",
       "email": "tlsteinberger167@gmail.com"
+    },
+    {
+      "name": "Nikhil Tilwalli",
+      "email": "ntilwalli@gmail.com"
```
staltz: 👍

staltz:
@ntilwalli can I ask you to rebase your branch with upstream's master?

If you haven't done this before,


- git fetch upstream (upstream points to this repo, origin should point to your fork on github)
- git checkout master
- git pull upstream master (this should pass with no conflicts at all, should be a simple fast forward)
- git push origin master (not important, but anyway...)
- git checkout your-pr-branch
- git rebase master (does the trick) 
- If things break during the rebase, you should fix the conflicts, then git add, then git rebase --continue
- When the rebase is done, git push origin your-pr-branch

TylorS:
@staltz This should wait until a new version of @most/create is released with fixes to the type definitions that are causing the tests to fail.

ntilwalli:
@staltz I rebased.  Yeah, the merge should be held off until the latest `@most/create` gets published to npm.  I forgot that I hacked the local type-definition for that lib to get this to build. I'll update the PR with the updated dependency version number when it gets released, then the PR can be merged.

TylorS:
Simply waiting for the npm build to be published for @most/create as it has been merged.

ntilwalli:
@Tylor cool.  Let me know when it gets published.

TylorS:
@ntilwalli create has been updated to v1.1.3 on npm

ntilwalli:
@TylorS @staltz I updated the version dependency.  Good to go.

TylorS:
@ntilwalli Thanks again for your work on this.

TylorS:
@ntilwalli Released in v4.0.1

Issue #355, opened 2016-07-17
Cannot cancel an XMLHttpRequest in @cycle/http


artfuldev:
With the addition of the runStreamAdapter, there is a default listener added, which takes away the ability to cancel XHRs.

staltz:
Thanks for reporting, seems serious.

artfuldev:
@staltz was not able to submit my entry to your flux-challenge because of this. Was breaking my head over why until @Hypnosphi found this.

Hypnosphi:
The problem is that there are some cases when we want requests to complete even if we don't listen to responses (some POST requests, for example). Maybe `makeHTTPDriver` should accept some `{lazy: true}` option which will indicate that we're going to listen to all the `response$` explicitly.

staltz:
I have a proposal for this.

```js
  let request$ = xs.of({
    url: 'http://localhost:8080/hello', // GET method by default
    category: 'hello',
    cancelWhen: xs.periodic(2000).take(1),
  });
```

Cancels the XHR after 2 seconds.

This would be a non-breaking change, which is good. What do people think?
Also @TylorS

eshansingh:
So cancelWhen takes a stream, that, when completed, cancels the XHR. Am I getting this?

staltz:
Takes a stream that when it emits `next` *or* `complete`, cancels the XHR.

eshansingh:
It is a really good idea, and it's even better that its non-breaking! :+1: from me.

Hypnosphi:
I don't think I can really know in advance what will be the reason of cancellation. Let's look at cycle HTTP docs:

```js
let httpResponse$ = httpResponse$$.flatten(); // flattens the metastream
  // the reason why we need to flatten in this API is that you
  // should choose which concurrency strategy to use.
  // Normal xstream flatten() has limited concurrency of 1, meaning that
  // the previous request will be canceled once the next request to the
  // same resource occurs.
  // To have full concurrency (no cancelling), use flattenConcurrently()
```

It's natural to expect here, that the bare fact that we stop listening to previous request leads to its cancellation. In fact this functionality seems to be [provided](https://github.com/cyclejs/cyclejs/blob/master/http/src/http-driver.ts#L100), but it's not in real, because of the driver's listener presence.
So how do you achieve the same functionality with `cancelWhen` field on the request object?

staltz:
Yeah, you're right, the driver's noop listener is a problem. So far we know these requirements for sure:

[...]

PR #388, opened 2016-08-09
fix(http): allow lazy (cancellable) requests

Hypnosphi:
ISSUES CLOSED: #355

Usage: `{url: 'http://example.com', lazy: true}`.

I have no idea how to add tests for request abortion, need help here.


whitecolor:
@Hypnosphi 
Does it all to make driver request lazy be default?

Maybe should be reverse `eager` option instead of `lazy`, what is more semantically expressive?
@staltz @TylorS

Hypnosphi:
The main idea is that non-breaking changes are generally better than breaking

whitecolor:
@Hypnosphi what breaking change are you talkin about?

Hypnosphi:
Changing default behaviour would break a lot

whitecolor:
I mean not changing default behaviour, but adding a driver option. By default all requests should be eager, but you may set them to be lazy with this option.

An another thing is just a word: `lazy` vs `eager` what is more semantically expressive.

staltz:
Thanks @Hypnosphi this looks good, and I agree for now it's better to introduce a simple non-breaking feature.

whitecolor:
@staltz are you ok with `lazy` name for this param?

And I'm not proposing any breaking change. I said that this param should be also added for the whole driver, but in case of `lazy` param it will be `false` by default, what breaking changes are here?

staltz:
Yeah I think `lazy` is a good name since it works as an opt-in.

whitecolor:
Ok, nice.

whitecolor:
Ok nice. I've used the same param for new version of `cycle-async-driver`, here is docs: https://github.com/whitecolor/cycle-async-driver#lazy-drivers-and-requests

whitecolor:
@staltz what about driver scope `lazy` param (false by default)?

staltz:
Let's discuss that one a bit more, since it's a larger feature. This PR was a very small addition, but fixed the bug for those needing a fix.

whitecolor:
Well it will just assume `lazy: true` for all the requests (as it was in the previous version of HTTP, driver) I see no problems here.

whitecolor:
@staltz though you maybe right, and default laziness may not be is not needed for HTTP driver case.

Hypnosphi:
I think it's doable now, the driver option, given that you always have an opt-out by apparently passing `lazy: false` in the request

Hypnosphi:
Now I think of a bit wider feature: driver accepts an options hash, which will be defaults for all the request. These can even include urlRoot which will be prepended to all the urls

staltz:
Can we discuss this in a new issue?

Issue #368, opened 2016-07-25
DOM Source to support document and body


staltz:
We could support `domSource.select('document')` and `domSource.select('body')` as terminal domSources (meaning we don't know what should happen with additional `select()` after that) because there are cases where `domSource.select('document').events('click')` would be useful and easy to support.

### If you want to help build this

**assignee: @raquelxmoss**

- [ ] Add test for `domSource.select('document').events('click')` and perform a programmatic click on any part of the DOM, then assert
- [ ] Add test for `domSource.select('body').events('click')` and perform a programmatic click on any part of the DOM, then assert 
- [ ] Implement the feature in MainDOMSource.ts, in the `select` method. Somehow `document` and `body` should have special treatment. One idea is that `select('document')` would return an instance of the class `DocumentDOMSource implements DOMSource` and `select('body')` would return an instance of `BodyDOMSource implements DOMSource`. `DocumentDOMSource.select` could return always `this` instance of `DocumentDOMSource`, since it doesn't make sense to progressively select on that type.

raquelxmoss:
If this isn't super urgent (which it doesn't seem to be), I'd like to take this on! I've been using the mouse driver for this, but if it's added to the core, we can probably deprecate parts of the mouse driver.

I haven't written typescript before, so this will be a good introduction to it.

staltz:
Thanks @raquelxmoss, that would be great! Go ahead 👍

wizardwerdna:
I'd be happy to tackle this is raquel doesn't have time for it.  In any case, I would propose a friendly amendment to the spec, to use `:document` and `:body` in lieu of the naked definitions (as we do with `:root`.

raquelxmoss:
Hey folks 😄 I'm working on it this weekend, but if I don't get it done I'll hand it over 🌈

eshansingh:
See PR #389

staltz:
Cycle DOM v12.1.0 released with this feature. Thank you @raquelxmoss !

PR #389, opened 2016-08-13
Add support for DOM.select('document') and DOM.select('body').

raquelxmoss:
Chained calls to select will be ignored if the first select in the chain
is `document` or `body`.

ISSUES CLOSED: #368


staltz:
Merged via commit https://github.com/cyclejs/cyclejs/commit/93bd3c45ba6556ecd2909a34f6e6582abbd2ce28. 
Many thanks @raquelxmoss !

PR #399, opened 2016-08-24
refactor(examples): refactor bmi-nested to use MVI pattern

craigmichaelmartin:
I refactored the bmi-nested example to use the [Model-View-Intent pattern](http://cycle.js.org/model-view-intent.html) during lunch to play around with cyclejs. I also moved `isolate` back to the component (which now [exports a fn which calls isolate](http://cycle.js.org/components.html#Should-I-always-call-isolate-manually)) and changed the sinks keys' name to [not use the stream convention](http://cycle.js.org/components.html#how-to-name-sources-and-sinks).

These changes were for my own learning, but figured I'd make a PR in case you would rather have the example follow this pattern. 


staltz:
Thank you, this looks ok

PR #411, opened 2016-08-30
Remove unnecessary export

craigmichaelmartin:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ node_modules
 npm-debug.log
 .DS_Store
 ignore/
+*.swo
+*.swp
--- a/xstream-run/src/index.ts
+++ b/xstream-run/src/index.ts
@@ -33,8 +33,8 @@ import XStreamAdapter from '@cycle/xstream-adapter';
  * Cycle.js program, cleaning up resources used.
  * @function run
  */
-export function run<Sources, Sinks>(main: (sources: Sources) => Sinks,
-                                    drivers: {[name: string]: Function}): DisposeFunction {
+function run<Sources, Sinks>(main: (sources: Sources) => Sinks,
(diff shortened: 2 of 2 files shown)


whitecolor:
wow are you sure it is unnecessary??

craigmichaelmartin:
Short answer: no I am not. It made good theoretical sense to me (and the tests passed!), but I may be stretching some assumptions. I have to run to an appt and then I'll double back on this, if it is still open. I'll read [this](https://medium.com/@kentcdodds/misunderstanding-es6-modules-upgrading-babel-tears-and-a-solution-ad2d5ab93ce0#.p5pdidsfd) and the [referenced](https://t.co/oEHVWplBJQ) [articles](http://www.2ality.com/2014/09/es6-modules-final.html) in the meantime.

staltz:
Thanks for the urge to contribute, it's really appreciated.

About the change itself, sorry, but it doesn't make any sense. We need `import Cycle from '@cycle/xstream-run'` and **also** import {run} from '@cycle/xstream-run'`.

If you want to contribute we desperately need help in many issues, check these marked "help wanted": https://github.com/cyclejs/cyclejs/labels/help%20wanted

craigmichaelmartin:
Hi @staltz. Thank you for your response and all your work! I apologize for the erroneous PR. To your comment: I am aware both are necessary. My faulty assumption, articulated in the referenced articles, was that default exported objects could be destructured in the import (It was intended that`run` would not need to be exported because it would be destructured from the default export `Cycle` which has it as a [property](https://github.com/cyclejs/cyclejs/blob/master/xstream-run/src/index.ts#L83)). 

As a friendly recommendation, I would avoid telling others their work doesn't make any sense :)

PR #420, opened 2016-09-03
fix(http): fix typings after removal of response$$

whitecolor:
@staltz 
Tests need to be rewritten in TS too.

- [ x] I ran `npm test` for the package I'm modifying
- [x ] I used `npm run commit` instead of `git commit`

Remove response$$ from interfaces and fix filter and select signatures


staltz:
Thanks @whitecolor

PR #421, opened 2016-09-03
fix(most-run): export run funcrion

Hypnosphi:

Allows `import {run} from 'most-run'` usage. Also includes corresponding test for all *-run packages to prevent things like #411.


staltz:
Thanks @Hypnosphi

Issue #409, opened 2016-08-30
Isolated DOM stops receiving events after parent element replaced


chromakode:
Hi!

**Code to reproduce the issue:**

Unfortunately I wasn't able to get webpackbin to load the latest version of `@cycle/dom` (kept hanging on downloading deps), so here's an old fashioned repo with a demo: https://github.com/chromakode/cyclejs-dom-isolate-repro

In this example, an isolated component is being inserted into a parent node with a changing key. When that parent node's key changes, the element's content is re-rendered. The isolated component stops receiving events after this occurs.

**Expected behavior:**

Clicking the first button should increment the count once per click.

**Actual behavior:**

The count only goes up to 2, after which click events are ignored.

**Comments**

I believe the issue lies in the behavior of `isolateModule`'s `destroy` hook:

https://github.com/cyclejs/cyclejs/blob/0126d2ee917d15d8e42b799b369944d1985e19b4/dom/src/isolateModule.ts#L103

In the debugger, I've observed that when the parent element is recreated, we first see the `create` hook run with the new element, and then `destroy` is called on the old one. This causes the `eventDelegators` list to be emptied, even though we just added a new valid element to it.

I believe we could fix this by filtering the `eventDelegators` list so we only clear out delegators associated with the removed element. We should probably do this for both the `destroy` and `remove` hooks. I'd be happy to write a PR for this if it's on the right track.

Thanks! :)

**Versions of packages used:**

```
  "dependencies": {
    "@cycle/dom": "^12.2.2",
    "@cycle/isolate": "^1.4.0",
    "@cycle/xstream-run": "^3.1.0",
    "snabbdom-jsx": "^0.3.1",
    "snabbdom": "chromakode/snabbdom#fix-attribute-removal",
    "xstream": "^6.2.0"
  },
  "devDependencies": {
    "babel-core": "^6.9.1",
    "babel-loader": "^6.2.4",
    "babel-preset-es2015": "^6.9.0",
    "babel-plugin-transform-react-jsx": "^6.8.0",
    "webpack": "^2.1.0-beta.20",
    "webpack-dev-server": "^2.1.0-beta.0"
  }
}
```

Hypnosphi:
What's the point of changing the key when the element stays semantically the same?

chromakode:
This is a bit of a contrived example. :smile:

[...]

PR #426, opened 2016-09-06
Fix isolated component events after parent is recreated

chromakode:
Here's a proposed fix to #409.

The added test case fails on master and passes w/ this new change. I'm reusing `isolatedElements` as a way to check whether the element has been replaced by a new instance (when the parent is recreated, the create hook seems to run first and then the old node gets destroyed).

### DIFF
--- a/dom/test/browser/isolation.js
+++ b/dom/test/browser/isolation.js
@@ -715,6 +715,46 @@ describe('isolation', function () {
     dispose = run();
   });
 
+  it('should handle events when parent is removed and re-added', done => {
+    let clicksCount = 0;
+
+    function Child(sources) {
+      sources.DOM.select('.foo').events('click').subscribe(() => clicksCount++);
+      return {
+        DOM: Rx.Observable.of(div('.foo', ['This is foo']))
+      };
+    }
+
+    function main(sources) {
+      const child = isolate(Child)(sources);
+      // change parent key, causing it to be recreated
+      const innerDOM$ = Rx.Observable.timer(0, 120).take(3)
+        .map(x => div({key: x}, [child.DOM]));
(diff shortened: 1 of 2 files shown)


dom/test/browser/isolation.js
```diff
@@ -715,6 +715,46 @@ describe('isolation', function () {
...
+    let dispose;
+    sources.DOM.select(':root').elements().skip(1).subscribe(function (root) {
+      setTimeout(() => {
+        const foo = root.querySelector('.foo');
+        if (!foo) return;
+        foo.click();
+      }, 0);
+    });
+    setTimeout(function(){
+      assert.strictEqual(clicksCount, 3);
```
staltz: Thanks for the PR. Help me understand something: why are we expecting it to be `3`? The foo element was clicked just once.

chromakode: Yeah, this is kind of tricky -- I was puzzled by it too. It's copied from the test case immediately above.

What's happening is that the `:root` `.elements()` stream is emitting an event each time the DOM is updated. The main `DOM` stream emits 3 times so we expect 3 clicks. I'd be happy to rewrite this in a more direct fashion if you'd like.

staltz: Ah, I see it now. How about you put `elements().skip(1).take(3)` just to make it clear the intent is to test for 3 elements?

chromakode: Good idea! Done.

chromakode:
I rebased to fix the first commit message -- changed the classification from "style" to "refactor".

staltz:
Great! Merged. Thanks a lot @chromakode

PR #428, opened 2016-09-08
fix(http): remove 'remember()' from response$$, not value-over-time

ntilwalli:

There is no reason to represent response$$ using a MemoryStream, it is not a value-over-time, HTTP responses are events

### DIFF
--- a/http/src/http-driver.ts
+++ b/http/src/http-driver.ts
@@ -148,8 +148,7 @@ function makeRequestInputToResponse$(runStreamAdapter: StreamAdapter) {
 export function makeHTTPDriver(): Function {
   function httpDriver(request$: Stream<RequestInput>, runSA: StreamAdapter, name: string): HTTPSource {
     let response$$ = request$
-      .map(makeRequestInputToResponse$(runSA))
-      .remember();
+      .map(makeRequestInputToResponse$(runSA));
     let httpSource = new MainHTTPSource(response$$, runSA, name, []);
     /* tslint:disable:no-empty */
     response$$.addListener({next: () => {}, error: () => {}, complete: () => {}});


staltz:
Hi @ntilwalli, thanks for helping out.

Because this was a low-cost and high-risk PR, I think we need to discuss carefully why we need or do not need this. It's low cost because it's a very small change, and seems to pass all tests. But it's high risk because I'm not sure if we need remember() or not, and if there is some corner case that justifies that we need it, then this change would break code for existing users.

I think it's a problem when I don't understand whether we need something or not, so I'd like to be sure that we either need it or don't need it.

The value over time argument is good, and we should remove remember. On the other hand going through the commits in Cycle HTTP, there was a point where we added "autoSubscribe"/"eager", which came with the addition of ReplaySubject. Why? I can't remember. But I believe I did an informed decision at that time. Maybe I forgot to add a test.

So, I'll think about it. And if anyone else is reading this, help us out to be sure on either A or B.

ntilwalli:
@staltz Understood. This came up as an issue for me while building a component.  In my app I happen to swap in a component for a different one. Both components uses the exact same category for the `select` filter. When the first component sends a request with that category and then the second component is swapped in, the new component receives a response for a request it never made since, upon connecting the HTTP source, the previous response from the older component is cached in the MemoryStream.  This fixes a bug in my opinion.

staltz:
Oh, I see. That's good input.

staltz:
Yeah, I fail to see why we would need `remember`. Thanks

whitecolor:
This will not work for example in such case: http://www.webpackbin.com/EJZK3wW2b

```
function Main({HTTP}) {    
  return {    
    HTTP: xs.of('http://sandbox.webpackbin.com'),
    log: HTTP.select().flatten()
  }
}
```
due to `HTTP.select()` is subscribed after `request$` inside the driver. I'm not sure which one is more correct.

PR #438, opened 2016-09-16
fix(rxjs-adapter): fixes isValidStream, broken by xstream v6.3

ntilwalli:

### DIFF
--- a/rxjs-adapter/src/index.ts
+++ b/rxjs-adapter/src/index.ts
@@ -39,6 +39,7 @@ const RxJSAdapter: StreamAdapter = {
   isValidStream(stream: any): boolean {
     return (
       typeof stream.subscribe === 'function' &&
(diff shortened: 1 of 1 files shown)


staltz:
Makes sense. Thanks @ntilwalli

staltz:
I'm restarting the Travis build to see if we have an actual test failing or not

staltz:
```js
  1) HTML Driver should output HTMLSource as an adapted stream:

      AssertionError: 'undefined' === 'function'

      + expected - actual

      -undefined

      +function
```

This test seems to be consistently failing. Is it happening because of xstream v6.3?

staltz:
> Is it happening because of xstream v6.3?

Yes

staltz:
I think Cycle DOM is breaking because it's expecting this PR to be released. :)

staltz:
Thanks @ntilwalli !

PR #440, opened 2016-09-18
fix(dom): enable restarting of event streams on isolated components

ntilwalli:

This is a fix for a bug I encountered which is replicated here: http://github.com/ntilwalli/isolateTest

When the DomTest component is isolated I was not able to restart the `itemMouseUp$`

The fix pulls the `hadIsolated_mutable` value out of the closure and into the stream state to ensure that restarted subscriptions cause the `hadIsolated_mutable` value to be reinitialized to `false` (when it wasn't being reinitialized, the remembered DOM was being filtered out for all restarts of the isolated events stream)

__Notes:__ the `isolateTest` uses a local copy of `rxjs-adapter` due to a bug introduced into `rxjs-adapter` a few days ago when `xstream` was updated.  The fix submitted in https://github.com/cyclejs/cyclejs/pull/438 is required and must be installed for that repo to properly replicate this bug.

One test fails which has nothing to do with this fix (it fails even without this update) and shows this output:
```
  1) HTML Driver should output HTMLSource as an adapted stream:

      AssertionError: 'undefined' === 'function'
      + expected - actual

      -undefined
      +function
      
      at Context.<anonymous> (test/node/html-render.js:85:12)
```

### DIFF
--- a/dom/package.json
+++ b/dom/package.json
@@ -53,7 +53,7 @@
     "babel-plugin-syntax-jsx": "^6.3.13",
     "babel-plugin-transform-react-jsx": "^6.4.0",
     "rx": "^4.1.0",
-    "rxjs": "^5.0.0-beta.8",
+    "rxjs": "^5.0.0-rc.1",
     "saucie": "^1.4.1",
     "simulant": "^0.2.2",
     "snabbdom-jsx": "^0.3.0",
--- a/dom/src/MainDOMSource.ts
+++ b/dom/src/MainDOMSource.ts
@@ -142,15 +142,17 @@ export class MainDOMSource implements DOMSource {
     const key = keyParts.join('~');
     const domSource = this;
     let rootElement$: Stream<Element>;
+
     if (scope) {
-      let hadIsolatedMutable = false;
       rootElement$ = this._rootElement$
-        .filter(rootElement => {
-          const hasIsolated = !!domSource._isolateModule.getIsolatedElement(scope);
-          const shouldPass = hasIsolated && !hadIsolatedMutable;
-          hadIsolatedMutable = hasIsolated;
-          return shouldPass;
-        });
+        .fold(function shouldPass(state, val) {
+            const hasIsolated = !!domSource._isolateModule.getIsolatedElement(scope);
+            const shouldPass = hasIsolated && !state.hadIsolated_mutable;
+            return {hadIsolated_mutable: hasIsolated, shouldPass: shouldPass, dom: val};
+        }, {hadIsolated_mutable: false, shouldPass: undefined, dom: undefined})
+        .drop(1)
+        .filter(x => x.shouldPass)
+        .map(x => x.dom);
     } else {
       rootElement$ = this._rootElement$.take(2);
     }
--- a/dom/test/browser/events.js
+++ b/dom/test/browser/events.js
@@ -4,7 +4,8 @@ let assert = require('assert');
 let Cycle = require('@cycle/rxjs-run').default;
 let CycleDOM = require('../../lib/index');
 let Rx = require('rxjs');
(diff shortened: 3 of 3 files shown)


staltz:
Thanks @ntilwalli, on the surface this looks reasonable. But I think since it looks like an important bug, we need to have some tests in place. Either me or you or Tylor could make the test in this PR, but anyway we need the test.

ntilwalli:
This being my first attempt at writing a test for `@cycle/dom`, I could use some guidance.  I wrote the below test, but it seems not to execute (it times out).  Can you quickly see why?
```
it('should allow restarting of event streams from isolated components', function (done) {
    function component() {
      return {
        DOM: Rx.Observable.of(button('.item', [`stuff`]))
      }  
    }

    function app() {
      return isolate(component)
    }

    function mouseevent(el, type) {
      // const ev = document.createEvent(`MouseEvent`)
      // ev.initMouseEvent(
      //   type,
      //   true /* bubble */, true /* cancelable */,
      //   window, null,
      //   0, 0, 0, 0, /* coordinates */
      //   false, false, false, false, /* modifier keys */
      //   0 /*left*/, null
      // )
      const ev = new MouseEvent(type)
      el.dispatchEvent(ev)
    }

    const {sinks, sources, run} = Cycle(app, {
      DOM: makeDOMDriver(createRenderTarget())
    });

    const itemMouseDown$ = sources.DOM.select('.item').events('mousedown')
    const itemMouseUp$ = sources.DOM.select('.item').events('mouseup')
    const itemMouseClick$ = itemMouseDown$
      .switchMap(down => {
        return itemMouseUp$.filter(up => down.target === up.target)
      })
    const switchSubject = new Rx.Subject()
    let count = 0
    switchSubject.switchMap(function () {
      return itemMouseClick$
    }).subscribe(ev => {
      assert.strictEqual(ev.type === 'mouseup')
      count++
      if (count === 2) {
        done()
      }
    })

    sources.DOM.select(':root').elements().skip(1).take(1).subscribe(root => {
      console.log(`Hello!!!!`)
      const clickable = root.querySelector('.item');
      setTimeout(() => mouseevent(clickable, 'mousedown'));
      setTimeout(() => mouseevent(clickable, 'mouseup'), 5);
      setTimeout(() => switchSubject.next(undefined), 10);
      setTimeout(() => mouseevent(clickable, 'mousedown'), 15);
      setTimeout(() => mouseevent(clickable, 'mouseup'), 20);
    });


    run();
  });
```

ntilwalli:
@staltz I posted an update which includes a test which is failing for a reason I don't understand.  This is my first foray into `@cycle/dom` tests. The fix works in the browser for my project so I suspect I've written an improper test, but it's also possible my proposed fix doesn't work.  Can you quickly tell if my test has an issue?  

As it stands the subscriber sees one event but not the second.

Also, the update mandates xstream 6.2 since 6.3 causes many failures.

staltz:
@ntilwalli just checking, is this ready for review? (I'm not saying it's not, just asking)

ntilwalli:
@staltz Yes, it's ready for "review",  not for merge yet.  I added a test but it fails due to a timeout, which means emissions aren't happening and I'm not clear on why.  I could use some guidance on what I might be doing wrong.  I'm suspect the failure comes because of how I've coded the test and not the fix...  Any thoughts?

ntilwalli:
@staltz This PR is ready to go.  I added failing test which passes with the new code. 

I determined I wasn't leaving enough delay between simulated clicks and that was causing the test to fail earlier. I'm using `setTimeout` in my test to simulate a serial set of clicks, dunno if there's a better way.

Travis/Sauce Labs is failing for some reason not having to do with the fix.

staltz:
Superb! Thanks @ntilwalli !

PR #442, opened 2016-09-27
fix(dom): add virtual DOM sanitation on `dispose()`

### LINKED ISSUE #263: Proper DOM cleanup while driver disposal
This problem is actual for such use cases as hot-reload of cycle app, or then cycle apps are pereodically instantiated while runtime.

In such scenarios when don't need running app anymore we dispose all sinks and sources, then probably cleanup HTML and replace with with new instance.

The problem is that while DOM driver disposal vdom tree is not cleaned up and thus `destroy` hooks on elements (widgets) do not work, this may cause some leaks and side effects.

To address this issue I curretly use use this as a solution:
```js
const _makeDOMDriver = (contaner, options) =>
  (dom$) => {
    let empty$ = new BehaviorSubject()
    empty$.onNext(false)
    let source = makeDOMDriver(contaner, options)(
      Observable.combineLatest(dom$, empty$, (dom, empty) => {
        return empty ? div('') /* empty dom */ : dom}
      )
    )
    let _dispose = source.dispose
    source.dispose = () => {
      empty$.onNext(true)
      _dispose.apply(source)
    }
    return source
  }
```

I wonder if something like this (vdom cleanup) should be done by DOM driver by default while disposal.

Frikki:
- [ ] I used `npm run commit` instead of `git commit`

When `dispose()` was called, the virtual DOM tree wasn’t cleaned properly;
thus, `destroy` hooks on elements (widgets) wouldn’t work and could cause
leaks and side effects.

With this fix, calling `dispose()` destroys the virtual DOM tree that was
created and cleans up all produced effects.

Closes #263

### DIFF
--- a/dom/src/MainDOMSource.ts
+++ b/dom/src/MainDOMSource.ts
@@ -69,6 +69,7 @@ function determineUseCapture(eventType: string, options: EventsFnOptions): boole
 
 export class MainDOMSource implements DOMSource {
   constructor(private _rootElement$: Stream<Element>,
+              private _sanitation$: Stream<{}>,
               private _runStreamAdapter: StreamAdapter,
               private _namespace: Array<string> = [],
               public _isolateModule: IsolateModule,
@@ -116,6 +117,7 @@ export class MainDOMSource implements DOMSource {
       this._namespace.concat(trimmedSelector);
     return new MainDOMSource(
       this._rootElement$,
+      this._sanitation$,
       this._runStreamAdapter,
       childNamespace,
       this._isolateModule,
@@ -204,6 +206,7 @@ export class MainDOMSource implements DOMSource {
   }
 
   dispose(): void {
+    this._sanitation$.shamefullySendNext('');
     this._isolateModule.reset();
   }
 
--- a/dom/package.json
+++ b/dom/package.json
@@ -14,6 +14,10 @@
     {
       "name": "Tylor Steinberger",
       "email": "tlsteinberger167@gmail.com"
+    },
+    {
+      "name": "Frederik Krautwald",
+      "email": "fkrautwald@gmail.com"
     }
   ],
   "keywords": [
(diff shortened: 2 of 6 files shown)


Frikki:
So CI fails in *cyclejs/history* as the record shows it also did four months ago. This failure has nothing to do with the provided fix.

dom/src/MainDOMSource.ts
```diff
@@ -116,6 +117,7 @@ export class MainDOMSource implements DOMSource {
       this._namespace.concat(trimmedSelector);
     return new MainDOMSource(
       this._rootElement$,
+      this._sanitation$,
```
Widdershin: I like the name sanitation, but I wonder if `cleanUp$` or `dispose$` might be more clear.

dom/src/MainDOMSource.ts
```diff
@@ -204,6 +206,7 @@ export class MainDOMSource implements DOMSource {
   }
 
   dispose(): void {
+    this._sanitation$.shamefullySendNext('');
```
Widdershin: Do we need to send an empty string through? Looks like you can call `shamefullySendNext` with no arguments. (Unless this is just to make TypeScript happy, ignore me if so).

dom/src/MainDOMSource.ts
Frikki: I’d like to rename when I do a larger refactoring process on a new branch.

Frikki:
@TylorS Could you merge this PR? All checks have passed.

staltz:
Anyway, thank you @Frikki for coming back and helping us :)

Widdershin:
@staltz I'm currently blocked on needing this for the cycle-restart xstream port, any chance it could be released as a patch? :smile: Aware the refactor is in progress but that seems maybe a few days away.

staltz:
@Widdershin Done. DOM v12.2.6 is out

PR #460, opened 2016-10-16
Renamed vdom$ to vtree$

hhariri:
The declared stream and the returned one weren't matching.


staltz:
Thank you @hhariri, makes sense.

Issue #471, opened 2016-11-12
Dom docs options wrong describtion


SuperManitu:
The README of the DOM Driver states:
```
'modules: array' contains additional Snabbdom modules
```
It should be something like this:
```
'modules: array' overrides the default Snabbdom modules
```
Talked with @TylorS about this

PR #472, opened 2016-11-12
Fix Option Text

SuperManitu:
See #471

### DIFF
--- a/dom/README.md
+++ b/dom/README.md
@@ -52,7 +52,7 @@ the app on the DOM.
 
 - `container: String|HTMLElement` the DOM selector for the element (or the element itself) to contain the rendering of the VTrees.
 - `options: DOMDriverOptions` an object with two optional fields: `transposition: boolean` enables/disables transposition of inner streams in
-the virtual DOM tree, `modules: array` contains additional Snabbdom modules.
+the virtual DOM tree, `modules: array` overrides the default Snabbdom modules.
 
 #### Return:
 


dom/README.md
```diff
@@ -52,7 +52,7 @@ the app on the DOM.
 
 - `container: String|HTMLElement` the DOM selector for the element (or the element itself) to contain the rendering of the VTrees.
 - `options: DOMDriverOptions` an object with two optional fields: `transposition: boolean` enables/disables transposition of inner streams in
-the virtual DOM tree, `modules: array` contains additional Snabbdom modules.
+the virtual DOM tree, `modules: array` overrides the default Snabbdom modules.
```
mightyiam: How about a link where I can see what the default Snabbdom modules are?
And it should be also clear, whether these defaults are in Cycle.js or upstream Snabbdom defaults.

TylorS: Good idea @mightyiam. The defaults are to Cycle.js. Snabbdom has no concept of any defaults at present. The defaults for Cycle.js are the `class`, `attrs`, `props`, and `style` modules as an FYI.

staltz:
Hi :)
Thanks for submitting this PR, we appreciate any kind of desire to help. Before merging this PR I need you to notice that this README.md is always generated from the source. So could you please make a PR where we update only the JSDoc comment in the source code? That will be reflected in the generated README.md

SuperManitu:
will do tomorrow

PR #477, opened 2016-11-16
fix(http): fix sensitivity to ordering of sinks

### LINKED ISSUE #476: Cycle run sensitive to order of sinks

```js
import Cycle from '@cycle/xstream-run';
import xs from 'xstream';
import flattenConcurrently from 'xstream/extra/flattenConcurrently';
import {div, label, input, hr, ul, li, a, makeDOMDriver} from '@cycle/dom';
import {makeHTTPDriver} from '@cycle/http';

function main(sources) {
  const getTodos = xs.of({
    url: 'http://jsonplaceholder.typicode.com/todos',
    category: 'todos',
    method: 'GET'
  })

  const todoList = sources.HTTP.select('todos')
    .compose(flattenConcurrently)
    .map(res => res.body)
    .debug()

  const vdom = todoList.map(todos => {
    return div(todos.map(todo => div(todo.title)))
  })

  return {
    HTTP: getTodos,
    DOM: vdom,
  }
}

Cycle.run(main, {
  DOM: makeDOMDriver('#main-container'),
  HTTP: makeHTTPDriver()
});
```

**Expected behavior:**

xstream `debug()` should have seen a value, the vdom too and therefor the DOM sink as well, and those items should have been displayed.

**Actual behavior:**

HTTP Source does not emit.

```js
"@cycle/dom": "^14.0.0",
"@cycle/http": "^11.2.0",
"@cycle/xstream-run": "^3.1.0",
"xstream": "^7.0.0"
```

- - -

Here's why:

```js
  return {
    HTTP: getTodos,
    DOM: vdom,
  }
```

does not work.

```js
  return {
    DOM: vdom,
    HTTP: getTodos,
  }
```

works.

**Rationale:** in Cycle.run replicateMany, it would subscribe to the HTTP sink first, and then to the DOM sink second, because of the ordering in the sinks object. That would mean the synchronous request would be consumed, but it wouldn't be available by the time we subscribe to the DOM sink, because response$$ is not a value-over-time (and we recently received a PR from @ntilwalli to make response$$ not be a value-over-time, because that would cause bugs).

staltz:

The solution here is to simply delay all requests by a little bit, so we never process requests synchronously. This is okay as long as requests are always asynchronous anyway.

ISSUES CLOSED: Closes #476.

### DIFF
--- a/http/src/http-driver.ts
+++ b/http/src/http-driver.ts
@@ -1,4 +1,5 @@
 import xs, {Stream, MemoryStream} from 'xstream';
+import delay from 'xstream/extra/delay';
 import {MainHTTPSource} from './MainHTTPSource';
 import {StreamAdapter, DriverFunction} from '@cycle/base';
 import XStreamAdapter from '@cycle/xstream-adapter';
@@ -154,6 +155,7 @@ export function makeHTTPDriver(): Function {
                       runSA: StreamAdapter,
                       name: string): HTTPSource {
     let response$$ = request$
+      .compose(delay(1))
       .map(makeRequestInputToResponse$(runSA));
     let httpSource = new MainHTTPSource(response$$, runSA, name, []);
     response$$.addListener({next: () => {}, error: () => {}, complete: () => {}});
--- a/http/test/common.js
+++ b/http/test/common.js
@@ -287,6 +287,40 @@ function run(uri) {
         output.run();
       }
     );
+
+    it('should not be sensitive to ordering of sinks (issue #476)', function(done) {
+      function main(sources) {
+        var request$ = Rx.Observable.of({
+          url: uri + '/hello',
+          method: 'GET',
+        });
+        var response$ = sources.HTTP.select()
+          .mergeAll()
+          .map(function (res) { return res.text; });
+
+        // Notice HTTP comes before Test here. This is crucial for this test.
+        return {
+          HTTP: request$,
+          Test: response$,
+        }
+      }
+      var testDriverExpected = ['Hello World'];
+
+      function testDriver(sink) {
+        sink.subscribe(function (x) {
+          assert.strictEqual(testDriverExpected.length, 1);
+          assert.strictEqual(x, testDriverExpected.shift());
+          assert.strictEqual(testDriverExpected.length, 0);
+          done();
+        });
+      }
+
+      var output = Cycle(main, {
+        HTTP: makeHTTPDriver(),
+        Test: testDriver,
+      });
+      output.run();
+    });
   });
 
   describe('isolateSource and isolateSink', function () {


staltz:
**Open for review**

In particular, a very important question we should ask is: is this change entirely backwards compatible? If not, we should label it as a breaking change.

@ntilwalli @TylorS @Frikki ?

whitecolor:
> The solution here is to simply delay all requests by a little bit, 

I think that from mathematical standpoint it is not a valid solution to delay all the requests, when this is needed only for requests that are made synchronously when function is executed.

staltz:
Where's the "math" to back that argument? 🙄

whitecolor:
When you making only necessary operations that means you logic is backed by mathematical evaluation. When you are making some operation just to "plug the hole" is just patching that potentially leads to bugs and unexpected behaviors. Timeout is JS is bad, they also mess with schedulers in stream realm.

As other proposal, more complex but more mathematically approved solution: driver function may contain additional property (like `streamAdapter`) function say `.callWhenCycleIsReady` (name is just to bring the meaning) that is used to inform the driver that cycle is ready (called when all internal subscriptions are made). So HTTP driver in this case will start send responses to response$$ only when cycle is ready.

ntilwalli:
I've actually experienced this issue a bunch and never considered it a bug. Whenever a component needs to send a sync http request in my app I add a zero delay directly to the HTTP sink for that component. I can see the "it's a bug" perspective since It does leak the wiring aspect of what cycle does but once I got that, it never bothered me.

@whitecolor what harm does adding a zero delay to the HTTP sink do in the run call? It's already an async call and pushing the request to the next eventloop turn is effectively no cost

ntilwalli:
That being said @whitecolor's solution could also be applied to other sinks which require synchronous emission which would be clean. I have no strong opinions either way

staltz:
> function say .callWhenCycleIsReady that is used to inform the driver that cycle is ready

@whitecolor even in that case, what if you have multiple drivers that have this property, which among those do you pick to run first? It wouldn't solve the problem because in the face of two such drivers, ordering their sinks incorrectly in the app may yield the same bugs.

What this fix does is guarantee that no request occurs synchronously, which is (1) a good guarantee against these types of bugs, (2) a reasonable choice since we know HTTP is inherently async. We could have chosen different ways of making sure it's asynchronous. For instance one method I tried was
```diff
-request$
+xs.periodic(1).take(1).mapTo(request$).flatten()
```
But that doesn't work since request$ is hot and it would still emit synchronously, and then we would still have the "too late, missed it" issue.

Another one I thought of doing is
```js
const firstReq$ = request$.take(1).delay(1);
const restReq$ = request$.drop(1);
```
But we don't have any guarantee that the first request will be synchronous. In fact we could have multiple synchronous requests, and this method wouldn't work.

Ideally we would just delay only synchronous requests, not delay the already-async requests. But on the other hand, how do you do that in practice?

whitecolor:
> In fact we could have multiple synchronous requests, and this method wouldn't work.

Right, because of this *remembering* the stream (that was removed some time ago) too didn't eliminate the issue.

>  what if you have multiple drivers that have this property, which among those do you pick to run first?

Solution proposed by me just may ensure the driver (if it needs it) that at least initial subscriptions inside `Main` function has been made and circuit is closed (in `replicateMany`). No need to care what is going on **inside other drivers**, they do not affect the circuit.

Widdershin:
Just wanted to weigh in, I updated cycle-restart to the latest version of @cycle/http and some of my core http tests started failing.

Manually applying this fix solves the problem.

I can provide more detail if it's useful but just wanted to say I'm looking forward to this patch landing.

staltz:
@Widdershin can you let me know if PR https://github.com/cyclejs/cyclejs/pull/484 would also fix the issue you had?

Widdershin:
@staltz #484 also fixes the issue 😄

PR #479, opened 2016-11-21
docs(dom): update README to include ownerTarget

FeliciousX:
The existence of `ownerTarget` is not very obvious to `cycle/dom` users. Hence the README is updated to reflect the existence of it.


staltz:
Thanks @FeliciousX !

Issue #412, opened 2016-08-31
DOM driver doesn't receive events from inside the shadow dom


tusharmath:
**Code to reproduce the issue:**

``` js

// DOM element needs to have a parent other wise snabbdom won't render
function createWrapperNode () {
  const wrapper = document.createElement('x-wrapper')
  const node = document.createElement('x-node')
  wrapper.appendChild(node)
  return {node, wrapper}
}

// Lifecycle method — http://www.html5rocks.com/en/tutorials/webcomponents/customelements/
function createdCallback () {
  const shadowRoot = this.createShadowRoot()
  const {parent, node} = createWrapperNode()
  this.__vNode = node
  shadowRoot.appendChild(parent)
  const drivers = {
    DOM: makeDOMDriver(this.__vNode)
  }
}
```

**Expected behavior:**

``` js
DOM.select('.some-selector').events('click') // should fire
```

**Actual behavior:**
No event is fired.

**Versions of packages used:**

``` json
    "@cycle/dom": "^12.2.2",
    "@cycle/rx-run": "^7.0.1",
    "@cycle/xstream-run": "^3.1.0",

```

Creating web components using cyclejs helps in having a consistent paradigm across the application. I have created a special driver to handle attribute changes but need help with the DOM driver.

tusharmath:
@staltz what are your thoughts on this?

tusharmath:
Sorry, this wasn't a real issue.

tusharmath:
I was able to reproduce this here — https://esnextb.in/?gist=5e6c955cb6072800b3a73f7abfe56fce

staltz:
Released in Cycle DOM v14.2 thanks to @ajmchambers.

PR #482, opened 2016-12-01
fix(dom): change EventDelegator bubble to search topElement instead of body

ajmchambers:

Searching the body will not find the element inside a DocumentFragment or shadowRoot, so the function ends up exiting early and the event is ignored. Added a test for a click inside a DocumentFragment, behaviour is similar to shadowRoot but has better browser support.

ISSUES CLOSED: #412

### DIFF
--- a/dom/test/browser/events.js
+++ b/dom/test/browser/events.js
@@ -340,6 +340,52 @@ describe('DOMSource.events()', function () {
     dispose = run();
   });
 
+  it('should catch bubbling events in a DocumentFragment (and shadowRoot)', function (done) {
+    function app() {
+      return {
+        DOM: Rx.Observable.of(div([
+          div('.clickable', 'Hello')
+        ]))
+      }
+    }
+
+    function click(el) {
+      const ev = document.createEvent(`MouseEvent`)
+      ev.initMouseEvent(
+        `click`,
+        true /* bubble */, true /* cancelable */,
+        window, null,
+        0, 0, 0, 0, /* coordinates */
+        false, false, false, false, /* modifier keys */
+        0 /*left*/, null
+      )
+      el.dispatchEvent(ev)
+    }
+
+    const fragment = document.createDocumentFragment();
+    const renderTarget = fragment.appendChild(document.createElement('div'));
+
+    const {sinks, sources, run} = Cycle(app, {
+      DOM: makeDOMDriver(renderTarget)
+    });
+
+    sources.DOM.select('.clickable').events('click')
+      .subscribe(ev => {
+        assert.strictEqual(ev.type, 'click');
+        assert.strictEqual(ev.target.tagName, 'DIV');
+        assert.strictEqual(ev.target.className, 'clickable');
+        assert.strictEqual(ev.target.textContent, 'Hello');
+        assert.strictEqual(ev.target.parentElement.parentNode instanceof DocumentFragment, true)
+        done();
+      });
+
+    sources.DOM.select(':root').elements().skip(1).take(1).subscribe(root => {
+      const clickable = root.querySelector('.clickable');
+      setTimeout(() => click(clickable));
+    });
+    run();
+  })
+
   it('should have currentTarget or ownerTarget pointed to the selected parent', function (done) {
     function app() {
       return {
--- a/dom/test/browser/events.js
+++ b/dom/test/browser/events.js
@@ -17,6 +17,33 @@ function createRenderTarget(id = null) {
   return element;
 }
 
+function testFragmentEvents() {
+  function getResult() {
+    let captures = false;
+    let bubbles = false;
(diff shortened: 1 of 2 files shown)


Widdershin:
The test is timing out in a bunch of browsers. I wonder why?

ajmchambers:
Hmm yea, I'll see if I can figure out what's going on.

staltz:
Thanks @ajmchambers for this PR. It makes sense. I'll add some comments inline.

dom/test/browser/events.js
```diff
@@ -340,6 +340,52 @@ describe('DOMSource.events()', function () {
...
+      ev.initMouseEvent(
+        `click`,
+        true /* bubble */, true /* cancelable */,
+        window, null,
+        0, 0, 0, 0, /* coordinates */
+        false, false, false, false, /* modifier keys */
+        0 /*left*/, null
+      )
+      el.dispatchEvent(ev)
+    }
```
staltz: We use just `element.click()` in other tests, so I think you can remove this helper function.

dom/test/browser/events.js
```diff
@@ -340,6 +340,52 @@ describe('DOMSource.events()', function () {
...
+        assert.strictEqual(ev.target.tagName, 'DIV');
+        assert.strictEqual(ev.target.className, 'clickable');
+        assert.strictEqual(ev.target.textContent, 'Hello');
+        assert.strictEqual(ev.target.parentElement.parentNode instanceof DocumentFragment, true)
+        done();
+      });
+
+    sources.DOM.select(':root').elements().skip(1).take(1).subscribe(root => {
+      const clickable = root.querySelector('.clickable');
+      setTimeout(() => click(clickable));
```
staltz: Maybe add more time here, like `setTimeout(() => ...., 80)`.

ajmchambers:
I've been playing around with Chrome 39 (one of the browsers that failed the test) to see if I can figure out whats going on.

It would appear (in Chrome 39 at least) that events aren't bubbling in detached dom.

```javascript
var parent = document.createElement('div')
var child = parent.appendChild(document.createElement('div'));
parent.addEventListener('e', function() {
  console.log('event bubbled to parent');
});
child.addEventListener('e', function() {
  console.log('child dispatched an event');
});
child.dispatchEvent(new CustomEvent('e', { bubbles: true }));
```

Logs both to the console in modern chrome and safari etc, but in chrome 39 it logs the initial child event but not the parent one.

mightyiam:
Congrats on debugging win. Is it out of curiosity or is there a reason to support Chrome 39 specifically?

ajmchambers:
Travis failed because the test I wrote failed in the following browsers:

- iPhone Safari 8.0
- iPhone Safari 9.0
- Chrome 30.0
- Chrome 39.0
- Safari 8.0
- Safari 9.1

So in Chrome 39 (at least) I think my test failed because the event didn't actually bubble - which is the point of my test in the first place. Not sure about the other browsers yet, may be the same issue or something else.

I'm pretty new to this in general, but is there a best practice for tests that you expect to fail in certain situations and not others? (or is that just a bad test)

I suppose you could do this:

```javascript
function eventsBubbleInDetachedDOM() {
  // checks and returns boolean
}

it('should catch bubbling events in a DocumentFragment (and shadowRoot)', function (done) {
  if(eventsBubbleInDetachedDOM()) {
    // run the test
    done();
  } else {
    done();
  }
});
```

or this:

```javascript
function eventsBubbleInDetachedDOM() {
  // checks and returns boolean
}

if(eventsBubbleInDetachedDOM()) {
  it('should catch bubbling events in a DocumentFragment (and shadowRoot)', function (done) {
    // run the test
    done();
  });
}
```

But it doesn't seem right having branching logic for tests.

Any ideas?

(Busy at the moment so won't be able to have a proper look at this again until later on tonight - but when I do get a chance I will make the requested changes and see what happens).

staltz:
You can do something like: https://github.com/cyclejs/cyclejs/blob/master/dom/test/browser/render.js#L213

But I'm curious what exactly don't those old browsers support.

ajmchambers:
Hopefully this fixes it: added a check to see if events will capture or bubble in DocumentFragment and added a test for useCapture. Will check back later. 

Btw should I also be using `npm run commit` when updating a pull request?

ajmchambers:
Ok, now I'm a little confused: dom tests passed ok, but history tests are failing somehow... could my change in the dom EventDelegator have affected history somehow?

Will have a look into it later on. If it was could maybe change my check to something like:

```javascript
if ( 
  !(
    document.body.contains(rawEvent.currentTarget as Node) ||     
    this.topElement.contains(rawEvent.currentTarget as Node)
  )
) {
  return;
} 
```

staltz:
Thanks for the PR!

PR #488, opened 2016-12-10
remove duplicate "of"

miguelmota:


staltz:
Hi, thank you for taking your time to send us this issue. We appreciate it, however I'd like to point out that our README.md is generated from source code, so the next time we generate the README, these changes would be overridden/removed. Can you submit these changes to be in the JSDoc comments of `dom/src/index.ts` instead?

PR #489, opened 2016-12-10
refactor(rxjs-adapter): scope down rxjs module importing

MidnightWonderer:

Hi there  
I noticed that my bundle is really large and find out that rxjs-adapter load the entire rxjs5 library.  
It would be nice to only bundle what necessary to the adapter.


staltz:
Thanks @MidnightWonderer !

staltz:
New version of RxJS adapter launched. This also will automatically improve new installs of rxjs-run

PR #495, opened 2016-12-14
chore(META): exclude .babelrc from NPM builds

goshakkk:
- [ ] I added new tests for the issue I fixed/built (n/a)
- [ ] I ran `npm test` for the package I'm modifying (n/a)

It's not needed anyway. It's also causing issues when used with React-Native packager, which processes node_modules with Babel

(https://github.com/reactjs/redux/issues/1033#issuecomment-156584084)

### DIFF
--- a/base/.npmignore
+++ b/base/.npmignore
@@ -2,3 +2,4 @@ ignore/
 .idea/
 *.sublime-project
 *.sublime-workspace
+.babelrc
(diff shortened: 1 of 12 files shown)


staltz:
Thanks!

staltz:
I'm releasing a new minor version for each of these packages.

PR #496, opened 2016-12-15
chore(META): add test-wait to dom package.json scripts

ajmchambers:

Three options: test (same behaviour as before, so npm-x-for-all-packages test does not pause), test-ci (same behaviour as before), additional test-wait which leaves testem in a state of 'Waiting for runners...'

NOTE: made pretest-browser, posttest-browser the main scripts that *-browser-ci and *-browser-wait both share and reordered to: test-browser, test-browser-ci, test-browser-wait


staltz:
Thanks!

PR #499, opened 2016-12-17
feat(base): snabbdom example using hooks

klaaz0r:
As someone new to cycle I missed an example for working with modules like chart.js and I noticed more people having the same issue, I hope this small example shows what you can do with snabbdom hooks. It is based of the custom-driver example.

- [ ] I ran `npm test` for the package I'm modifying


staltz:
Hi, thank you for taking your time to send us this issue. We appreciate it, however this example is between Cycle.js and snabbdom. It's mostly a demo of how snabbdom works. I think it would be better to post this is a standalone example in another repository, and then you can list it in https://github.com/cyclejs-community/awesome-cyclejs.

klaaz0r:
He @staltz I am working on a dashboard containing more graphs and visualisation,  will make it standalone and see if I can move it to the cyclejs-community, keep up the awesome work with cycle!

PR #502, opened 2016-12-29
feat(http): added request object to http err response

Steelfish:
The request object is added to the superagent response object on non error statuscodes. On error the request object is unavailable in the error's response object. This pull request fixes that.

### DIFF
--- a/http/src/http-driver.ts
+++ b/http/src/http-driver.ts
@@ -89,6 +89,7 @@ export function createResponse$(reqInput: RequestInput): Stream<Response> {
         }
         this.request.end((err: any, res: Response) => {
           if (err) {
+            err.response.request = reqOptions;
             listener.error(err);
           } else {
             res.request = reqOptions;
--- a/http/test/common.js
+++ b/http/test/common.js
@@ -19,7 +19,7 @@ function run(uri) {
     this.timeout(8000);
 
     it('should throw when request stream emits neither string nor object',
-      function(done) {
+      function (done) {
         function main(sources) {
           return {
             HTTP: Rx.Observable.of(123)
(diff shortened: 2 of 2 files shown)


Widdershin:
@Steelfish it would be good if there was an automated test for this, to prevent any accidental regressions in future. I'm happy to help if you run into any issues writing a test.

Steelfish:
I added a test based on the response$ onError test. My habit to push format document seems to have added some spaces to various functions as well. Let me know if I should make any more changes.

whitecolor:
Generally I (as an author of PR that added `request` object to `response`) think it is not very good pattern to include request in the response, because it is already available in the metastream. Yes this simplifies a little bit things, but creates ambiguity (how request object can be accessed), so I would even remove `request` from `response`.

Steelfish:
I am currently using the request object on the response$ to keep track of outstanding requests. If there is another way to do this, I would love to hear it.

staltz:
I somewhat agree with @whitecolor, but we already have `res.request = reqOptions;` for non-error responses, so we could have that same also for the error case. This was actually originally your code: https://github.com/whitecolor/cycle-http-driver/commit/fef21c817622ea545ef2b56b225ec99826f0c9d0 so I'm curious what were the original ideas behind that PR https://github.com/cyclejs/http/pull/41.

whitecolor:
@staltz Idea was just easier access to corresponding request while handling response.

Though currently I don't use it at all because `http` driver is kind of `task` driver that just gets task request performs some async task action, and reports back with resolved or rejected result. So in general case (for other kind of task based drivers) we may not embed `request` into `response`, so that is way I consider it is inappropriate to embed it in `http` driver response too.

staltz:
It would be a breaking change to remove `response.request`, and one that wouldn't help people: those who are not using it won't notice, but those who are using it will need to refactor considerably. I'm favorable to merge @Steelfish's PR.

whitecolor:
As just a legacy it is ok.

`const addRequest = r$ => r$.map(r => Object.assgin(r, request: r$.request)`

`HTTP.select().map(addRequest)`

staltz:
Merged! Thanks @Steelfish !

PR #513, opened 2017-01-30
refactor(jsonp): rewrite jsonp driver for unified

Widdershin:
Now using the new global adapt call, and TypeScript.

### DIFF
--- a/jsonp/package.json
+++ b/jsonp/package.json
@@ -10,24 +10,22 @@
     "url": "https://github.com/cyclejs/jsonp"
   },
   "dependencies": {
-    "@cycle/xstream-adapter": "3.x.x",
+    "@cycle/run": "^1.0.0-rc.7",
     "jsonp": "0.2.0"
   },
   "peerDependencies": {
     "xstream": "*"
   },
   "devDependencies": {
-    "eslint": "1.0.0",
-    "eslint-config-cycle": "3.2.0",
-    "eslint-plugin-cycle": "1.0.2",
-    "eslint-plugin-no-class": "0.1.0",
-    "xstream": "8.x.x"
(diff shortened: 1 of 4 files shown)


Widdershin:
I tested this against the `autocomplete-search` example and it works 😄

staltz:
Awesome. Looks great.

There are some things to add, but I'll do that myself after merging.
One thing I was reminded is that we need to make `run` a hard dependency for all official drivers, like you did in this PR. Other drivers right now have `run` in devDependencies. This is because originally we had `@cycle/base` in devDependencies just to pull its TypeScript types, hence not needed for runtime.

So, `run` as a hard dependency is an action point in the big Unified todo list.

jsonp/package.json
```diff
@@ -10,24 +10,22 @@
...
-    "xstream": "8.x.x"
+    "xstream": "10.x.x"
   },
   "browserify-shim": {
     "xstream": "global:xstream"
   },
   "scripts": {
-    "lint": "eslint src/",
+    "compile": "../node_modules/.bin/tsc",
+    "prepublish": "npm run compile",
```
staltz: I'll do this myself, but notice how we call the script `lib`, not `compile`, in other packages.

staltz:
Also, let's use `"@cycle/run": "*"` as the dependency, because if the user already has `@cycle/run` in his package.json, we want to use exactly the same version of `run`, because we need to have the same `adapt` module. Otherwise we will have two different `run` versions installed, which means different state, and a bug.

Let's aim to not at all change the `adapt` API in `run`, so that `"*"` won't be a problem. If we do make a breaking change, than we need to change `"*"` of these packages to something else.

PR #515, opened 2017-02-02
Switch to karma tests for history driver in unified

TylorS:
Adds karma configuration necessary to run browser tests.
Adds some basic SauceLabs configurations,

Still needs to be configured for all browsers desired. 
@staltz Do you think you could pick that aspect up from here?


staltz:
So far so good! I'll continue from here :)

PR #516, opened 2017-02-02
PR for Cycle Unified

staltz:
Not yet ready.
This pull request was created early so that we can run tests in CI.


dom/test/node/html-render.ts
```diff
@@ -88,6 +89,7 @@ describe('HTML Driver', function () {
     let {sinks, sources, run} = setup(app, {
       html: makeHTMLDriver((html: string) => {}),
     });
+    console.log(sources.html.elements());
```
TylorS: Probably want to remove this log?

PR #517, opened 2017-02-02
refactor(dom): simplify implementation of mockDOMSource.select() and .events()

Widdershin:

Just happened to be looking at the `mockDOMSource` code and it seemed like it could be improved.

I feel like there's something I've missed and it was done this way originally for a reason.

The tests pass with this change. I haven't benchmarked it, but it should be more performant (`O(1)` instead of `O(n)`).


staltz:
Seems correct to me. Thanks @Widdershin !

PR #532, opened 2017-02-26
Use HTTPS for jsonplaceholder endpoints

yangshun:

The [get random user example](https://cycle.js.org/basic-examples.html#basic-examples-http-requests) doesn't seem to be working because https://cycle.js.org/ is served over HTTPS but hitting the HTTP version of jsonplaceholder.typicode.com.

This PR replaces `http://jsonplaceholder.typicode.com` to `https://jsonplaceholder.typicode.com`. Tested in `examples/basic-examples.html` and the HTTPS version works.


staltz:
Thanks!

PR #548, opened 2017-03-11
ResponseStream to extend Stream<response>

### LINKED ISSUE #547: Shouldn't ResponseStream extend Stream?
Here is [`ResponseStream`](https://github.com/cyclejs/cyclejs/blob/c2bbccec2bf4990abd9b1220ff1acfd49154f01a/http/src/interfaces.ts#L38-L40).

Since it is a stream, perhaps it should extend xstream's `Stream`?

I'm trying to provide a `ResponseStream` to xstream's `combine` and I'm getting this:

```
Argument of type 'ResponseStream' is not assignable to parameter of type 'Stream<any>'. Property '_prod' is missing in type 'ResponseStream'.
```

mightyiam:
And reorganize those types a little bit.

Closes #547.

### DIFF
--- a/http/src/http-driver.ts
+++ b/http/src/http-driver.ts
@@ -9,6 +9,7 @@ import {
   RequestOptions,
   RequestInput,
   Response,
+  ResponseMemoryStream,
 } from './interfaces';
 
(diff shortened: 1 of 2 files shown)


staltz:
Let's discuss #547 before merging this.

Issue #560, opened 2017-03-20
Sinks typing for `run` issue with interface


whitecolor:
```ts
interface MainSources {}

interface MainSinks {
  log: Stream<string>
}

const Main = (sources: MainSources): MainSinks => {
  return {
    log: xs.of('message')
  }
}

run(Main, { // is not ok
  log: () => ({})
})
```
The error is:

```
Argument of type '(sources: MainSources) => MainSinks' is not assignable to parameter of type '(sources: MainSources) => Sinks'.
  Type 'MainSinks' is not assignable to type 'Sinks'.
    Index signature is missing in type 'MainSinks'.
```
This is because of https://github.com/cyclejs/cyclejs/blob/master/run/src/index.ts#L276 

It is due to MainSink should satisfy `{[name: string]: FantasyObservable}` signature.

[We need to use `type` instead of `interface`](https://github.com/Microsoft/TypeScript/issues/14736#issuecomment-287817665) Though I alway before used `interface`s everywhere.

```ts
type MainSinks = {
  log: Stream<string>
}
```

But it is also possible to make it work with an `interface`, need to modify run typings, replace:

```ts
export function run<So extends Sources, Si extends Sinks>(
```
with
```ts
export function run<So extends Sources, Si extends FantasySink<Si>>(
```
where
```ts
type FantasySink<Si> = {[S in keyof Si]: FantasyObservable}
```

@staltz what do you think?

staltz:
> But it is also possible to make it work with an interface, need to modify run typings, replace:

If it's possible to make it work with an interface, without bringing breaking changes, then great, I agree we could change it. A PR for this would be nice. The tricky corner cases to be careful with are https://github.com/cyclejs/cyclejs/blob/87228167c22751a7e8aa3f2892891f9b74d2fe52/run/test/index.ts#L57-L116

In general I'll try to teach using `type` instead of `interface`, though.

staltz:
run v3.1 will be released with this. Thanks whitecolor

PR #562, opened 2017-03-24
Run sinks interface typing

whitecolor:
Fixes https://github.com/cyclejs/cyclejs/issues/560

- [x ] I added new tests for the issue I fixed/built
- [ x] I ran `npm test` for the package I'm modifying
- [ ] I used `npm run commit` instead of `git commit`

@staltz 
Sorry forget proper commit message, but also you maybe want to fix naming fo added `FantasyMap<Si>`

### DIFF
--- a/history/karma.conf.js
+++ b/history/karma.conf.js
@@ -38,13 +38,13 @@ module.exports = function (config) {
       version: '9.1',
     },
 
-    // TODO: fix this!
-    // SL_android_5_1: {
-    //   base: 'SauceLabs',
-    //   browserName: 'android',
-    //   platform: 'Linux',
-    //   version: '5.1',
-    // },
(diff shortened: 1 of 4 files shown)


staltz:
@whitecolor thanks for the PR.

I'll make some fixes/refactor after I merge. Overall, when the code change is not a breaking change, we should add tests, not *modify* a test. Because modifying a test is usually a sign of a breaking change.

Thanks

whitecolor:
I just didn't get your test well, and thought that it lacks Sink, and modification of the test would test also this change.

staltz:
The lacking Sink is on purpose, because in most cases it is better to leave sinks inferred, since they are obvious from the return type, but expected inputs can't be inferred. Of course typing the sinks separately should be allowed.

PR #565, opened 2017-03-26
docs(META): add typescript scaffold instructions

jvanbruegge:
As the one-fits-all flavor is the only typescript flavor that is compatible with caa v3

- [x] I used `yarn run commit` instead of `git commit`


staltz:
Hey, could you do the PR against `docs/content/getting-started.md` instead?

jvanbruegge:
Changed

staltz:
Thanks!

Issue #567, opened 2017-03-27
Add mockHTTPSource helper


Widdershin:
It would be nice if the HTTP driver provided a helper for mocking the HTTP source, comparable to mockDOMSource or mockTimeSource.

The API could be comparable to mockDOMSource.

Maybe:

```js
const response$ = xs.of({data: {some: 'thing'}});

const HTTP = mockHTTPSource({
  'categoryName': xs.of(response$)
});
```

staltz:
Of course!

jarlah:
Need help with this one? I have done some initial work, just creating the mock class. If i can work on it, is there any tips for how it can be best done? I have looked brielfy at the other mock source implementations.

Widdershin:
If you push up what you have and make a pull request, I'll review it and help you get it merged. 😸

jarlah:
@Widdershin its practically nothing atm. I am not too familiar with cyclejs internals yet, so ill have to poke around a little bit.

Widdershin:
Well I'm happy to answer any questions you have. I would model it after the
mockDOMSource :)

jarlah:
@Widdershin mockDomSource takes a nested config that can contain (selector) keys that contain other nested configs. This is like that because we want to use it like this:

```
    const userEvents = mockDOMSource({
      '.foo': {
        'click': Observable.of(135),
      },
      '.bar': {
        'scroll': Observable.of(2),
      },
    });
```

with the http mock source, you provided an initial example of:

```
const response$ = xs.of({data: {some: 'thing'}});

const HTTP = mockHTTPSource({
  'categoryName': xs.of(response$)
});
```

Therefore i make an assumption that the mock config for http source translates to something in the lines of:

```
export type MockConfig = {
  [categoryName: string]: Observable<Observable>;
};
```

where the signature ```Observable<Observable>```` obviously have to be something else. But given the initial usage example this is a somewhat stable assumption.

So the question is, is the assumption for the type signature of MockConfig correct, or should it be something else?

Widdershin:
@jarlah one thing to consider is that you can chain select calls, `HTTP.select('foo').select('bar')`. So ideally the config would be a tree.

Widdershin:
@jarlah need any help with this?

jarlah:
@Widdershin It boiled away. I think I had the gist of it. Its easter now, ill see if I can wrap up my thoughts and get somewhere. But yes, I may need some help in the end I think.

[...]

PR #571, opened 2017-03-30
feat(http): adding a skeleton for a new mock http source helper

jarlah:
As suggested in issue #567 there is a need for mocking the http source.


Widdershin:
@jarlah very nice start. The `.select()` implementation looks good.

Do you want to have a go at adding some tests next?

Have a try and I can give you some examples if you get stuck.

Also seems like we need some way of specifying responses that don't have a category, maybe we could have a special key like `:root` at the top level. Not sure what the best way to handle that is, any thoughts?

jarlah:
@Widdershin yes, you might be right about the :root suggestion. But given the dynamic nature of type signatures in typescript (and flow also), its possible to say that MockConfig is either X, Y or Z. The problem with this of course, is that you need to do runtime type checks to find out what the heck you are working with. But, I am actua|lly more worried about the api and usage, so I would suggest saying ```type MockConfig = Responses | { [categoryName?]: Responses | MockConfig }``` but i am new here so I am not sure if this is something one would let slip through ;)

jarlah:
@Widdershin btw i made a test for empty stream. But I need some input on how to make a test for some "actual" response streams. Because the stream that is passed in from the test should match the type signature ```Stream<MemoryStream<Response> & ResponseStream>```

jarlah:
```/bin/sh: 3: Syntax error: "done" unexpected (expecting "do")```
cant see that I have introduced something that could cause this to fail travis

**Comment by Widdershin** [comment]
I think the build failure is unrelated, seems to be an issue from staltz moving the config over to makefiles. I think he's on break for Easter, I'm sure he'll sort it out when he's back online.

I see the dilemma around the stream passed from the test matching `Stream<MemoryStream<Response> & ResponseStream>`. Personally it seems like it would be unreasonable to ask the user to specify all of the attributes involved in a `SuperAgentResponse`.

I think we can follow the precedent from `mockDOMSource`, which is to be permissive about what is actually in the streams. We are not forced to ensure that it is a `Stream<Event>`, it just has to be a stream.

So maybe we should try `export type Responses = Stream<any>` or `export type Responses = FantasyObservable`. We also want the user to optionally be able to set the request, so maybe make a type that is `Stream<any>` + optional request.

With those more permissive types, we can write a test like this:

```js
  it('supports categorized streams', (done) => {
    const response : any = {
      text: 'Hello world'
    }

    const response$ = xs.of(response);

    const userEvents = mockHTTPSource({
      'foo': xs.of(response$)
    });

    userEvents.select('foo').flatten().take(1).subscribe({
      next: (response: any) => {
        assert.equal(response.text, 'Hello world');
      },
      error: (err: any) => done(err),
      complete: () => done(),
    });
  });
```

Do you reckon that's a viable approach?

http/test/node/mock-http-source.ts
```diff
@@ -0,0 +1,21 @@
+import 'mocha';
+import * as assert from 'assert';
+import { mockHTTPSource } from '../../lib/index';
```
Widdershin: We probably want to change this to import from `../../src/index`, it should be okay since mocha is configured to compile any typescript files it finds as it goes.

Widdershin:
Fyi @jarlah I fixed CI so if you pull in master you can run the tests on Travis now.

jarlah:
@Widdershin thnx. I'll dive into it the coming days. If it hadn't been for all this work related stuff... :laughing:

jarlah:
@Widdershin I have now updated my branch and added your exampel test, and changed  the Respones type to be more open. We can disccuss your point about specifying request

Widdershin:
@jarlah sorry for my lack of followup on this, I must have seen your comment and then forgot about it.

This implementation is looking good. I had a thought about how we could possibly get the types all nice as well, let me know what you think:

We could create helpers to create `Response` and `Request` objects.

Change this:
```diff
-    const response: any = {
-      text: 'Hello world',
-    };
+    const response = mockResponse({
+      text: 'Hello world',
+    });
```

And for requests:

```js
const request = mockRequest({
  url: '/anything',
  method: 'POST'
}

const responseToRequest = mockResponse({
  data: {},
  request
}
```

jarlah:
Someone else needs to pick this up. Need to get comfortable with cyclejs before go headlong into opensource pull requests like this one.

Widdershin:
I'll pick this up.

PR #576, opened 2017-04-14
Use @cycle/time in examples

### LINKED ISSUE #575: Update examples to use @cycle/time
Now that @cycle/time is quite stable, we should update the examples to use it.

I plan on doing this, opening an issue so I can log against it for OpenCollective.

Widdershin:
Closes #575.

I have left the examples that use tween intact, as this is blocked on https://github.com/cyclejs/time/issues/13, which still requires a bit of discussion.

I have made the commits separately for each example I updated so that interested parties could see each one individually, but it might make sense to squash it all together upon merge.

I have manually tested that all examples work the same. 😸

### DIFF
--- a/examples/animated-letters/package.json
+++ b/examples/animated-letters/package.json
@@ -5,11 +5,12 @@
   "author": "Andre Staltz",
   "license": "MIT",
   "dependencies": {
+    "@cycle/dom": "15.0.0-rc.2",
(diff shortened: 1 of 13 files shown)


staltz:
Perfect! Reviewed and looks great

mightyiam:
Wow, thanks, @Widdershin .

PR #586, opened 2017-04-25
Documentation: Fix typo. Add missing backtick

jameslawson:
mockDOMSource missing closing bracket. Has a visible effect - the following sentence is shown in <code>.


staltz:
Hi @jameslawson, thanks for being involved in helping the framework. :)
The commit was for the HTML file, but we actually generate that automatically from source code. Could I ask you to do that change in `dom/src/index.ts` instead? I understand it's confusing that we have the HTML file too, and we're considering writing notes or warnings on how to avoid this problem in the future.

staltz:
Thanks!

Issue #584, opened 2017-04-24
Travis failing with new Makefile setup


staltz:
Apparently my macOS Makefile might differ from what Makefile Travis CI has.

whitecolor:
There where docker helps, or authoring scripts in js/ts, js scripts are much more flexible and less error prone (esp in case of TS) though seem at first to be more verbose in shell related tasks.

staltz:
PR for Makefile => JS would be nice.

Widdershin:
Shall we open another issue to get a genuinely cross platform solution? It would be nice to accommodate Windows users, even though the Windows subsystem for Linux exists it seems a lot of Windows devs aren't using it.

staltz:
We can, but it's a nice to have type of thing

Widdershin:
👍

PR #593, opened 2017-04-28
chore(META): fix makefile on linux to fix build

Widdershin:
Add new scripts to run tests for all packages

ISSUES CLOSED: #584

Let's see if this works. In the longer term we should probably find a fully cross platform solution but hopefully this will support OSX and Linux.

### DIFF
--- a/.scripts/pre-check-release.sh
+++ b/.scripts/pre-check-release.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # This script checks if the Travis build for the latest commit has passed,
 # and if it has, then it runs check-release. Otherwise it tells the status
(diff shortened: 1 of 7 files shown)


Widdershin:
Okay, got this passing, and I also verified that it fails.

Had to change the Makefile a bit to get it to report failures, turned out using `;` to chain commands was dropping the exit codes. It pretty much boils down to the difference between `false; true` and `false && true`.

I also changed it to run the tests before linting so that a lint error won't prevent the tests from running. The rationale here is that tests are more important than lint errors, so we should address test failures first and style issues second. (Ideally a contributor will run local tests anyway so this should not be very consequential).

@staltz this is ready to go, pending your review.

staltz:
Looks good! Let's wait for Travis, and I'll also run this on my mac and see that it all works

staltz:
Works on mac :+1:.

Widdershin:
Failure was sauce labs timing out, serves as a useful demonstration that it reports failures 😄

staltz:
Red is the new green.

PR #597, opened 2017-05-03
Added Streams selection to API docs 

SteveALee:
I hope this helps. It is not a complete PR though.

I couldn't figure out all the details for these for the docs. Let me know and I'll address them.

Also the other API docs have the JSDocs appended at `#API` which does not apply to this new page. I could not figure out how that is done in order to ensure it was disabled.

### DIFF
--- a/docs/.scripts/make-api-index.js
+++ b/docs/.scripts/make-api-index.js
@@ -13,6 +13,7 @@ var outputStr = template({
   content: content,
   premenu: [],
   postmenu: [
+    {title: 'Streams', link: 'api/streams.html'},
     {title: 'Cycle Run', link: 'api/run.html'},
     {title: 'Cycle RxJS Run', link: 'api/rxjs-run.html'},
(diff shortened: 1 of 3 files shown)


staltz:
Hi, thank you for taking your time to send us this PR. The `api` section is only for "API references" like listing the functions of a package. I'm not sure what you meant with "Streams selection" but in Getting Started we already explain how you can choose which stream library. What was your vision for this?

SteveALee:
@staltz  Ah, it seems I should have made that clearer.

a) we discussed adding links to the Stream docs - this could also be done in the individual Run pages
b) I feel that as streams are such a significant part of Cycle usage and used in the APIs that for completeness the supported streams should be mentioned and signposted in the reference section. Also for on boarding it makes the Ref section more complete, even if introduction section was not recently read.

Looks like I forgot the links which were the original motivation :(

If you feel this is useful then I'll add the links. If not then no matter.

Issue #583, opened 2017-04-24
Documentation for cycle/history


FeliciousX:
Lots of newcomers get stuck trying to use `cycle/history` or some basic routing. I was one of them.

Also, it's quite hard to understand and use `@cycle/history` just by looking the current documentation. I think it may need more examples.

I hacked some examples of usages [here](https://github.com/FeliciousX/cycle-history-docs) during the weekend. Would want feedback on how to improve on it and maybe put it as part of the official docs?

----------

### Proposed changes on the official docs:
- [x] Reorder `captureClicks` and place it after `makeHashHistoryDriver` ( first they need to understand `historyDriver` first before they can understand `captureClicks`'s use ) (#600)
- [x] Add example for `makeHistoryDriver` (#599)
- [x] Add simple example to the repo
- [ ] Add advanced example, making use of different actions that cycle/history supports

mightyiam:
My switch-path-router (not yet published) is aimed at becoming the de-facto
cycle.js router. The history driver is not a router, by itself. It is not
published yet because of an issue I've reported in switch-path that I
didn't get a response for, yet (@staltz ?).

staltz:
Hi :)
Thanks @FeliciousX. I think I would be glad to merge a PR for docs on cycle/history. To make it easier to review, I suggest making several small PRs instead of a huge one.

FeliciousX:
@mightyiam and an improved docs on history driver could help people understand better the inner workings of `switch-path-router` too! Instead of making it look like magic.

That's how I felt last time when I was using `cyclic-router` without fully understanding `cycle/history`

@staltz ok! Will make several small PRs into the docs

mightyiam:
@FeliciousX I had a similar experience. One good piece of doc is all it takes, sometimes. How about you make a PR and I promise to review it, as well.

FeliciousX:
will do it after working hours :D thanks in advance! @mightyiam

ajmchambers:
Hey all, think this is a great idea. I've recently been trying to figure this stuff out myself - so its all still kind of fresh in my mind (had a look at your examples too which were helpful).

Reading through the docs there were a couple of things that weren't very clear to me, though to be fair I usually prefer looking at examples rather than the api doc's themselves.

## makeHistoryDriver(options)

Maybe it was just me but initally I wasn't sure whether @cycle/history also required history itself.

## captureClicks(driver)

```javascript
const drivers = {
  history: captureClicks(makeHistoryDriver())
}
```

[...]

PR #599, opened 2017-05-07
docs(history): adds small example on makeHistoryDriver

FeliciousX:

since I'm modifying the docs, no test were ran.

Also, `yarn run commit` does not seem to work but I followed as closely as I can on the commit standards.

This is part of a series of small PRs for #583


staltz:
Good, thanks @FeliciousX !

PR #600, opened 2017-05-07
docs(history): re-order capture click to be after makeHashHistoryDriver

FeliciousX:

This PR is part of a series of small PRs for #583 

Moving `captureClicks` below might seem trivial but it makes more sense for people to understand the history driver first before knowing that you can use `captureClicks` to wrap around it. Especially after we add code examples.


staltz:
Thanks!

PR #601, opened 2017-05-07
DOM element helpers typings

whitecolor:
I want to open this PR (will contain multiple commits), for making better typings of DOM helpers based on @TylorS's https://github.com/TylorS/mostly-dom/blob/master/src/types/HtmlProperties.ts

@staltz please review initial changes, and say if you ok with the proposal


staltz:
Great initiative @whitecolor ! Keep on going with this please :)

Eventually it would be good to put this in snabbdom, **but** Cycle.js has a unique hyperscript helper, so it makes sense to do what you're doing now.

I'd also recommend writing some tests (they are in TypeScript nowadays) before implementing everything. It doesn't need to be full TDD, but just a few tests so we know what is the goal API that we want to get.

PR #603, opened 2017-05-12
feat(dom): add preventDefault option

jvanbruegge:
This would remove impure `preventDefault` calls inside the app. Usage:
```javascript
const event$ = sources.DOM.select('.myform').events('submit', { preventDefault: true });

//currently
const event$ = sources.DOM.select('.myform').events('submit')
    .map(ev => {
        ev.preventDefault();
        return ev;
    });
```

### DIFF
--- a/dom/src/BodyDOMSource.ts
+++ b/dom/src/BodyDOMSource.ts
@@ -27,6 +27,12 @@ export class BodyDOMSource implements DOMSource {
     } else {
       stream = fromEvent(document.body, eventType);
     }
+    if (options && options.preventDefault) {
+      stream = stream.map(ev => {
+        ev.preventDefault();
+        return ev;
+      });
+    }
     const out: DevToolEnabledSource & Stream<Event> = adapt(stream);
     out._isCycleSource = this._name;
     return out;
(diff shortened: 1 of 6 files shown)


staltz:
Hey Jan :)
Thanks for submitting this PR. Before merging it, though, I noticed that it's the simplest solution, it is likely we would refactor it later on (for instance, I can see how we would put the preventDefault call inside the EventDelegator or the fromEvent producer), for performance or for less use of tricky mutation. In that future scenario where this refactor happens, it would be important to make sure that the code still works as desired. So we need tests for this one. Usually I'm not strict on asking for tests for everything, but I think this is one of those cases where we need it. The code inside MainDOMSource.events() is very tricky and likely to be refactored.

jvanbruegge:
How would I test this?

Widdershin:
You could check that `event.defaultPrevented` is true when`preventDefault` is true and false otherwise.

https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented

jvanbruegge:
The CI fails because it was not able to start the iOS browser, the new tests works as expected

jvanbruegge:
Triggering a new build worked :)

staltz:
Great, thanks for the PR @jvanbruegge !

aronallen:
This is perfect for cycle/sandbox :)

PR #604, opened 2017-05-13
Enable to handle binary response

maskinoshita:
Add feature to handle binary response to cycle/http.

- [x] I ran `make test http` for the package http I'm modifying

### DIFF
--- a/http/.testem.json
+++ b/http/.testem.json
@@ -16,7 +16,8 @@
     "/json": { "target": "http://localhost:8070" },
     "/querystring": { "target": "http://localhost:8070" },
     "/error": { "target": "http://localhost:8070" },
-    "/delete": { "target": "http://localhost:8070" }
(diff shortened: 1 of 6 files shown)


staltz:
Good PR. Thanks @maskinoshita !

whitecolor:
Shouldn't this option be called [`responseType`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType)  not just `binary`?

staltz:
Good point @whitecolor. Because we haven't yet released this PR as a new version, this is still possible to change without a breaking change version. I think I agree with you, but I'll leave this up for discussion for a while.

maskinoshita:
I had not considered a useful naming.
I agree with you, @whitecolor. Thanks.

staltz:
We got 3 people agreeing on the name `responseType`. I'll do the rename and release a new version.

PR #605, opened 2017-05-13
docs(history): adds example for captureClicks

FeliciousX:
- [x] I used make commit instead of git commit

Part of #583


staltz:
Great!

Issue #65, opened 2015-01-11
Option to prevent default in domSource.select().events()


grozen:
Cycle allows turning virtually any DOM event into a stream to be consumed by intents. However, for some events this proves pointless as the event's default handling is disruptive.

For example, suppose I have a form and would like to receive a stream of submissions on it, I could do something such as:

``` javascript
Cycle.h('form#search', {onsubmit: 'searchSubmits$'}, [
  Cycle.h('input#q', {autocomplete: 'off', name: 'query', type: 'text', placeholder: 'Add term'})
])
```

While this would indeed create a stream, the form would still actually navigate to its target upon submission, thus denying me the ability to make sensible use of the stream.

What I would like to have (and will gladly make a reality) is something like this:

``` javascript
Cycle.h('form#search', {onsubmit: 'searchSubmits$!'}, [
  Cycle.h('input#q', {autocomplete: 'off', name: 'query', type: 'text', placeholder: 'Add term'})
])
```

Which will basically retain the same behavior, except it will also call `preventDefault` for the submit event, allowing use of the stream.

I am, of course, entirely open to suggestions on how to actually signify this behavior. Being a Ruby buff I am all for adding bangs to the end of things to signify anything drastic, but this is not necessarily the best way.

staltz:
Yes this should be fixed. The default Cycle behavior should be to preventDefault when using onsubmit: 'searchSubmits$' (without bang).
 
I am curious if there is any case at all where we want to allow the event's default instead of preventing it.
 
Hence, I'll fix this soon.

grozen:
Actually, the very first example [here](https://developer.mozilla.org/en/docs/Web/API/event.preventDefault) seems to fit the bill. People may still want their checkboxes to toggle even if the clicks are turned into a stream, no?

staltz:
Good point. Can you experiment whether the following approach solves your problem?

```javascript
var View = Cycle.createView(function () {
  return {vtree$: h('form#search', {onsubmit: 'searchSubmits$'})};
});

var Intent = Cycle.createIntent(function (view) {
  return {submitForm$: view.get('searchSubmits$')
    .map(function (ev) { ev.preventDefault(); return ev; })
  };
});
```

In order words, manually putting `preventDefault` in the Intent.

staltz:
@grozen I'm getting confident that you can just put the `ev.preventDefault()` in your Intents. Here is an example I put together: http://jsfiddle.net/staltz/wdxs67x9/2/

[...]

PR #606, opened 2017-05-15
feat(dom): allow predicate function for preventDefault

jvanbruegge:
ISSUES CLOSED: #65

A follow up for #603 

### DIFF
--- a/dom/src/BodyDOMSource.ts
+++ b/dom/src/BodyDOMSource.ts
@@ -1,7 +1,7 @@
 import xs, {Stream, MemoryStream} from 'xstream';
 import {adapt} from '@cycle/run/lib/adapt';
 import {DevToolEnabledSource} from '@cycle/run';
-import {DOMSource, EventsFnOptions} from './DOMSource';
+import {DOMSource, EventsFnOptions, Predicate} from './DOMSource';
 import {fromEvent} from './fromEvent';
 
 export class BodyDOMSource implements DOMSource {
@@ -27,9 +27,15 @@ export class BodyDOMSource implements DOMSource {
     } else {
       stream = fromEvent(document.body, eventType);
     }
-    if (options && options.preventDefault) {
+    if (options && typeof options.preventDefault !== 'undefined') {
+      let cond : Predicate = options.preventDefault as Predicate;
+      if (typeof options.preventDefault === 'boolean') {
+        cond = (() => options.preventDefault) as Predicate;
+      }
       stream = stream.map(ev => {
-        ev.preventDefault();
+        if (cond(ev)) {
+          ev.preventDefault();
+        }
         return ev;
       });
     }
--- a/dom/src/DOMSource.ts
+++ b/dom/src/DOMSource.ts
@@ -1,7 +1,8 @@
 import {Stream, MemoryStream} from 'xstream';
+export type Predicate = (e : any) => boolean;
 export interface EventsFnOptions {
   useCapture?: boolean;
-  preventDefault?: boolean;
+  preventDefault?: boolean | Predicate;
 }
 
 export interface DOMSource {
--- a/dom/src/DocumentDOMSource.ts
+++ b/dom/src/DocumentDOMSource.ts
@@ -1,7 +1,7 @@
 import xs, {Stream, MemoryStream} from 'xstream';
 import {adapt} from '@cycle/run/lib/adapt';
 import {DevToolEnabledSource} from '@cycle/run';
-import {DOMSource, EventsFnOptions} from './DOMSource';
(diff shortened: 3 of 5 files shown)


jvanbruegge:
It look like a lot of the tests are run multiple times for the same browser. Is that a timeout issue?

jvanbruegge:
Yes, I also noticed it locally, the tests are restarted quite often

staltz:
@jvanbruegge, thanks.
In this PR, before merging, let's try to get these things right:

- Avoid repeating tricky code
- Avoid closures that may tax performance (e.g. creating `cond` and using that in the stream `map`)

The 3 nested ifs are tricky code, specially with mutation too. I think it would be reasonable to move that code into `fromEvent` and/or EventDelegator. You have the chance of avoiding the map operation if the predicate is applied directly in the xstream Producer (or the xstream "Subject" in the EventDelegator).

aronallen:
Hi,

This is perhaps a separate PR, but I need something very similar for cycle/sandbox. preventDefault: true would work, but we can't pass functions between threads. It could be great if we could pass an object instead, that would compare the values against a set. like 
```
preventDefault: [{button: [0, 1]}] // button is 0 or 1
preventDefault: [{button: [0, 1]}, {keyCode: 13}] // button is 0 or 1 OR keyCode is 13
preventDefault: [{button: [0, 1], keyCode: 13}] // button is 0 or 1 AND keyCode is 13
```

jvanbruegge:
@aronallen You could wrap the DOM driver, taking the object and converting to a function in the main thread. I thinks it's a very specific use case

aronallen:
@jvanbruegge, yes it is very specific. 
The API here is very similar to what I imagined for cycle/sandbox, so it is great to see it is getting official support. We should also consider having this for stopPropagation, and stopImmediatePropagation.
It is quite easy to convert things in the DOM driver, but having this function would prevent WebWorkers from using any Cycle component, if there is a function declared in a sub component, it wouldn't transfer to the main thread.

staltz:
I'm happy this discussion popped up, because so far what we have in `master` in compatible with the use case that Aron describes. The function makes a difference when moving across boundaries. Overall we have to be aware of the format of data, because functions aren't naively serializable in JSON.

Aron, I'm with you that the use case should be supported, but the array API is kind of weird.

aronallen:
@staltz I would also prefer we found a better solution. The array / object thing was just an attempt to describe the condition with data, but I am open to other approaches. The way I thought about the API would be that it accepted one or a set of conditions, and conditions would be a key value pair, of the key, and the possible values, or a single value.

```
preventDefault : { keyCode : 13 } // a very simple use case
preventDefault : { keyCode : 13, shiftKey : false } 
preventDefault : { keyCode : [13, 10] } // prevent default on these keys
preventDefault : [{keyCode : [13, 10]}, { button : [0] }] // either left click, or key is 10 or 13.
```

jvanbruegge:
maybe something like ramda's `path` function:
`preventDefault: { test: ['target', 'dataset', 'someKey'], equals: 'something' }`
`preventDefault: { test: ['target', 'dataset', 'size'], greaterThan: 0 }`
`preventDefault: { test: ['target', 'dataset', 'someKey'], in: ['a', 'b', 'c'] }`

```
preventDefault: { test: ['target', 'dataset', 'index'], all: {
    greaterThan: 0,
    any: {
        lessThan: 100,
        greaterThan: 1000
     }
}}
```
This would map to `ev => ev.target.dataset.index > 0 && (ev.target.dataset.index < 100 || ev.taget.dataset.index > 1000)`

jvanbruegge:
So basicly:
```typescript
interface Comparators {
    equals? : any;
    notEquals? : any;
    greaterThan? : any;
    lessThan? : any;
    greaterOrEqual? : any;
    lessOrEqual? : any;

    in? : any[];

    all? : Comparator;
    any? : Comparator;
}
type PreventDefaultOptions = Comparator & { test: string[] };
```

staltz:
That would work. That said, I'm curious why do we really need conditional preventDefault in the first place? I went back through the discussions and it seems the use case was raised up by `@`abaco just as a "what if?". I think we need to ask ourselves what are the real-world-looking use cases for this feature, and whether it is worth solving, and also how much would it cost, i.e. the impact it has on performance or code size or code complexity.

jvanbruegge:
I dont know, I need only the boolean

staltz:
Okay, I would suggest keeping this PR around, then open an issue about *conditional preventDefault*, label it priority maybe, and go on to solve more important things.

jvanbruegge:
:+1:

jvanbruegge:
Closing this is favor of #610 (without functions) and #611

PR #610, opened 2017-05-17
refactor(dom): preventDefault does not manipulate stream

grozen:
These are most of the changes I was originally going to submit for #65 before #603 beat me to the punch. The main difference is that this does not manipulate the event streams themselves. This also incorporates support for the `preventDefault` option being a function, which was supposedly part of #603 as well before being scrapped for some reason. Not sure these are even still necessary, but who am I to judge?

I suspect some things I did might be slightly bold, so I'd love to hear your opinion.

### DIFF
--- a/dom/src/BodyDOMSource.ts
+++ b/dom/src/BodyDOMSource.ts
@@ -22,17 +22,9 @@ export class BodyDOMSource implements DOMSource {
 
   public events(eventType: string, options: EventsFnOptions = {}): Stream<Event> {
     let stream: Stream<Event>;
-    if (options && typeof options.useCapture === 'boolean') {
-      stream = fromEvent(document.body, eventType, options.useCapture);
-    } else {
-      stream = fromEvent(document.body, eventType);
-    }
-    if (options && options.preventDefault) {
-      stream = stream.map(ev => {
-        ev.preventDefault();
-        return ev;
-      });
-    }
+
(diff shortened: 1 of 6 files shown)


grozen:
Tests are running indefinitely, and the last time that happens a form submission was the culprit. I'll change the preventDefault tests to not use a form later today.

grozen:
That seems to have done the trick.

staltz:
Hi @grozen, this is great! Partially great. :D
Can you check the discussion in #606? We decided to post-pone preventDefault as a function because it's too complicated for WebWorker use cases, and doesn't seem so important.

That said, the refactor here to move the logic to fromEvent and EventDelegator is very welcomed. As well as the tweaks to tests.

grozen:
Huh... I did see it and was certain I posted a comment over there. I'll gladly yank out all the preventDefault as a function bits. Would you rather I clobber this with a force push or make a new one?

grozen:
Had some time to toss into this so I just clobbered everything. I'll pseudo-hang-on to a copy of the version that used functions just in case.

jvanbruegge:
Close, reopen

grozen:
Closing to rerun tests

staltz:
Looks good. Can I merge?

grozen:
Of course =D

staltz:
Merged it is.

PR #612, opened 2017-05-19
Fix wrong VNodeData typings

jvanbruegge:
The same issue as here: https://github.com/snabbdom/snabbdom/pull/296


staltz:
Thanks @jvanbruegge

Issue #611, opened 2017-05-19
Conditional preventDefault [title edited] 


jvanbruegge:
With the current `@cycle/dom` version, you can specify an option to let the driver call `preventDefault()` for you (keeping side effects outside of your app).
In #65 was the idea of a predicate function that could be used to conditionally `preventDefault`. The question is now, is that needed? Does anyone have a valid use case for that type of behavior? For more information about how this would look like, see https://github.com/cyclejs/cyclejs/pull/606#issuecomment-301862166

fuunnx:
I think the case of the conditional `keyDown` event is the most common:
I would want to call `preventDefault()` when the user presses the `enter` key in a form input, but not when he is typing.

staltz:
Very good point, and indeed, we even have an example where we preventDefault on `tab` pressed: 

- https://github.com/cyclejs/cyclejs/blob/91d56057bf5b1cafe6d56d554ff0da3c92092f65/examples/autocomplete-search/src/app.js#L132-L133
- https://github.com/cyclejs/cyclejs/blob/91d56057bf5b1cafe6d56d554ff0da3c92092f65/examples/autocomplete-search/src/app.js#L281

So I would say, yes, we certainly need conditional preventDefault.

staltz:
We just need to discuss what kind of API to use to provide this feature. I feel like a predicate function would follow the *principle of least surprise*. But a function might break use cases that @aronallen raised, like when crossing execution boundaries (WebWorker). The other option discussed in #606 was inspired by Ramda's path, as @jvanbruegge mentioned. Maybe there is yet another possible API?

jvanbruegge:
As supporting a predicate function is rather easy with `typeof X === 'function'` vs `typeof X === `object`, we could support both too, so the possible input would be:
```ts
type PreventDefaultOpt = boolean | (a : any) => boolean | ComparatorObj:
```

staltz:
Good idea! It doesn't need to be either-or. :) 👍

aronallen:
@jvanbruegge maybe we should keep the comparator object simple for now.
simply a key value comparison where every predicate must strictly equal
e.g.

```
preventDefault: {keyCode: 13}
```

staltz:
The only thing I'd worry about a polymorphic signature is the code size needed to implement it. I think 1kB or 2kB is the maximum before having to consider putting it in an external plugin or something.

jvanbruegge:
[...]

PR #615, opened 2017-05-24
Predicate fn

jvanbruegge:
As discussed in #611 

### DIFF
--- a/dom/src/fromEvent.ts
+++ b/dom/src/fromEvent.ts
@@ -1,6 +1,6 @@
 import {Stream, Producer, Listener} from 'xstream';
 
-export type Predicate = (ev : any) => boolean;
+export type Predicate = (ev: any) => boolean;
 export type PreventDefaultOpt = boolean | Predicate | Comparator;
 export type Comparator = { [key: string]: any };
 
@@ -34,7 +34,7 @@ export function fromEvent(
   );
 }
 
-export function preventDefaultConditional(event: any, preventDefault: PreventDefaultOpt) : void {
+export function preventDefaultConditional(event: any, preventDefault: PreventDefaultOpt): void {
   if (preventDefault) {
     if (typeof preventDefault === 'boolean') {
       event.preventDefault();
--- a/dom/src/fromEvent.ts
+++ b/dom/src/fromEvent.ts
@@ -46,17 +46,21 @@ export function preventDefaultConditional(
         event.preventDefault();
       }
     } else if (typeof preventDefault === 'object') {
-      const matchObject: (m: {}, o: {}) => boolean = (matcher, obj) =>
-        Object.keys(matcher).reduce(
-          (acc: boolean, k: string) =>
-            acc &&
-            (typeof matcher[k] === 'object' && typeof obj[k] === 'object'
-              ? matchObject(matcher[k], obj[k])
-              : matcher[k] === obj[k]),
-          true,
-        );
+      const match: (m: any, o: any) => boolean = (matcher, obj) => {
+        const isArray = Array.isArray(matcher);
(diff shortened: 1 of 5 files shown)


staltz:
Looking good, @jvanbruegge! I'll make some comments inline

dom/test/browser/src/events.ts
```diff
@@ -873,4 +873,70 @@ describe('DOMSource.events()', function () {
     run();
   });
 
+  it('should allow preventing default event behavior with function', function (done) {
```
staltz: Would be good to have tests also for the *opposite* case. If the implementation is `if (true) ev.preventDefault()` then these tests would pass.

staltz:
Good. I'll also ask @aronallen's opinion on the object-based API because it's better to get it right in the beginning than to issue a breaking change later on, requiring people to migrate code.

aronallen:
@staltz looks fine for now, the ramda approach may cover us better in the future, but I think this is much simpler and will work in 99% of cases.

dom/src/fromEvent.ts
```diff
@@ -30,3 +33,38 @@ export function fromEvent(
...
+): void {
+  if (preventDefault) {
+    if (typeof preventDefault === 'boolean') {
+      event.preventDefault();
+    } else if (typeof preventDefault === 'function') {
+      if (preventDefault(event)) {
+        event.preventDefault();
+      }
+    } else if (typeof preventDefault === 'object') {
+      const matchObject: (m: {}, o: {}) => boolean = (matcher, obj) =>
```
staltz: What's the purpose of the matchObject local function if it's only called in one place (just below its declaration)?

jvanbruegge: it's used twice, most noticably recursive here: https://github.com/cyclejs/cyclejs/pull/615/files/49431069da3dab1759b22c0253cf35949d2c69dc#diff-4974c1e7e3b97190ed437163ebe8d3adR54

jvanbruegge: Without this recursive matcher, something like `{ ev: { target: { value: 'something' }}}` would not be possible

aronallen: would this work for arrays as well?
This is a bit of a thought example, but something like:
```javascript
{ target : { classList: ['a', 'b'] } }
```

jvanbruegge: No, currently it does only an equality check, but i will update this PR later to allow this

jvanbruegge: @aronallen Done

jvanbruegge:
@staltz is this ready now?

staltz:
Yes, looks all good to merge. 11 days ago when you asked this I was quite busy with conferences. If you don't mind, I'll rebase/squash some commits because `feat` commits will be shown in the changelog and also some of them used `feat(DOM)` instead of `feat(dom)`.

jvanbruegge:
Yeah, go ahead

staltz:
Actually, @jvanbruegge, I think there are still two things to consider: tests for the array part of `match`, I think it was working correctly for arrays even before the commit "feat(DOM): Check arrays on recursive matcher", because `Object.keys(['a', 'b'])` gives `["0", "1"]`. Also, `match` function can be pulled out and become a top-level function. This will actually have better performance because currently every call to `preventDefaultConditional()` will create a **new** `match` function, and functions are objects, so we're spending memory here.

jvanbruegge:
A problem is that I lost my Laptop, so I have no idea when I could change that.

staltz:
Okay, I'll do it then, no stress. And sad to hear you lost your laptop :(
Hope you recover it, or the data, or a new laptop.

staltz:
See PR #678 as a continuation to this.

PR #626, opened 2017-06-02
chore(META): add each package's minified, gzipped size

craigmichaelmartin:

If this is desired, a script could be added to report these numbers.
EDIT: Or it looks like shields.io has a badge for this. I don't have any familiarity these badges but would be happy to look into it.


staltz:
Hi @craigmichaelmartin , thank you for taking your time to send us the PR. We appreciate it, however hardcoding the actual sizes isn't future proof since we will be changing the code size of each library quite often. I'd recommend looking into another approach such as the one used here https://github.com/staltz/xstream/commit/6ead197a967832b58dc2eb084e933433c218bed5

PR #629, opened 2017-06-03
refactor(META): apply prettier to all TypeScript files

staltz:
It's easier to review this commit only: https://github.com/cyclejs/cyclejs/commit/612fce757802f4faad31190e580b7588468f40be

The rest of this PR is the actual usage of prettier on the whole codebase.

Reviews please? :)


dom/src/IsolateModule.ts
```diff
@@ -23,14 +23,18 @@ export class IsolateModule {
 
   constructor() {
     this.elementsByFullScope = new MapPolyfill<string, Element>();
-    this.delegatorsByFullScope = new MapPolyfill<string, Array<EventDelegator>>();
+    this.delegatorsByFullScope = new MapPolyfill<
+      string,
+      Array<EventDelegator>
+    >();
```
jvanbruegge: :/

staltz:
@mightyiam that should work with this workflow.

staltz:
@jvanbruegge I know there are cases where even with a limit of 90 lines you can get line of code that would be beautiful on one line of 93 characters, but not when split into multiple lines. So there's no convenient number we can choose (unless it's like 150, but then it starts looking ugly in other ways). I'd say we just use prettier as a way of not discussing code style ever, so we don't spend time with this. If it looks a bit ugly, whatever, at least it will be consistent and always follow some logic, not any person's style.

staltz:
And there is something special about `80` itself:

> We strongly recommend against using more than 80 columns. Prettier works by craming as much content as possible until it reaches the limit, which happens to work well for 80 columns but makes lines that are very crowded. When a bigger column count is used in styleguides, it usually means that code is allowed to go beyond 80 columns, but not to make every single line go there, like prettier would do.

https://github.com/prettier/prettier/

jvanbruegge:
yeah, you are right

staltz:
You think it's okay to merge?

jvanbruegge:
Yes

staltz:
Well, CI fails for a reason not related to this code change, so I'll merge. Thanks @jvanbruegge for reviewing :)

craigmichaelmartin:
@staltz Awesome! Looks great! FYI, it seems like the workflow @mightyiam was describing (trying to commit only a portion of a file) won't work with this setup, since prettier runs for the full file. I can think of a few ways to mitigate this (only running `prettier --write` if `prettier -l` fails - so contributors auto-running prettier in their IDE wont run into this; or only run `prettier -l` in `make commit` and leave running prettier writing to the contributor altogether), but do you see this as an issue?

workflow

staltz:
Hi @craigmichaelmartin. :)

If a file is already prettified before a contributor has touched it, when they modify the file, prettier will run for the whole file, but only those parts that were modified should be altered, right? I don't think we will hit a case where prettier will format code unrelated to the staged changes.

Did I understand you correctly?

craigmichaelmartin:
Thanks for your quick response and effort to understand me. Sorry for not being clear :)

> I don't think we will hit a case where prettier will format code unrelated to the staged changes.

It seems like prettier doesn't format code *just* related to the staged changes, it formats the entire file and then `git add`s the entire file. The files returned from `$(git diff --cached --name-only --diff-filter=ACM | grep '\.ts\?$' | tr '\n' ' ')` are not the --cached (staged) versions, but the "real" files, if that makes sense.

EG:

```diff
import {Stream, MemoryStream} from 'xstream';
export interface EventsFnOptions {
  useCapture?: boolean;
  preventDefault?: boolean;
+ someProp?: any;
}

export interface DOMSource {
  select<S extends DOMSource>(selector: string): S;
- elements(): MemoryStream<Document | Element | Array<Element> | string>;
+ elements(): MemoryStream<Document | Element | Array<any> | string>;
  events(eventType: string, options?: EventsFnOptions): Stream<Event>;
}
```

If I just add the first change, only it is staged and everything is good so far. When I go to `make commit` it though, prettier runs over the entire file, and then the entire file is `git add`ed (including the second change, which I intentionally did not stage and which will implicitly be staged and committed without my realizing).

Does that help articulate articulate my concern?

staltz:
Now I see, thanks. It's a rather important (or should I say: useful) issue to fix. If it's possible to only run prettier on the staged lines, then let's do that. Otherwise, we could run `prettier -l` and block committing before all files are prettified.

craigmichaelmartin:
Yeah. I don't know an elegant way to do this with git. A hacky way though, lol, could be:

```bash
# Get the staged files
tsfiles=$(git diff --cached --name-only --diff-filter=ACM | grep '\.ts\?$' | tr '\n' ' ')

# Do the commit
$(BINDIR)/git-cz

# Stash other files
git stash

# Run prettier
for f in $tsfiles
do
  echo "Formatting $f with prettier"
  $(dirname $0)/../node_modules/.bin/prettier --write \
  --single-quote --no-bracket-spacing --trailing-comma=all $f;
  git add $f
done

# Amend the commit to include prettier changes
echo "Updating commit with prettified files"
git commit --amend --no-edit

# Bring back stashed files
git stash pop
```

I wonder what sort of problems would surface if you went this route 🙃 

Or, as you say, just block the commit if running `prettier -l` over the staged files fails. The downside of this is that `prettier -l` may fail for an aspect of the file the dev was hacking on but specifically didn't want to stage/commit.

If you know the direction you'd like to go, I'd be happy to make a PR.

staltz:
Amending sounds like a good hack. I think I would merge such PR, just make sure to test it a couple of times, since we don't have automated tests for this. :)

craigmichaelmartin:
Sounds good :) I'll do it tonight!

Issue #619, opened 2017-05-29
Replace uglifyjs with Google Closure Compiler in all packages


staltz:
We've done this in xstream recently and it was a good idea. Uglifyjs isn't being maintained that well and GCC is potent. Seems like an easy win.

jvanbruegge:
I was confused by gcc :D
Just so used to writing C/C++

maximilianmikus:
I would like to try to help with this issue. Have not committed to cyclejs before, but I figure this might be a good starting point.

staltz:
Yes, it would be a good starting point, if you can navigate through our custom monorepo setup. Thank you for volunteering! Here's a reference https://github.com/staltz/xstream/commit/ed6f79360eb527d38a25aa26efc01c7970961a0c

maximilianmikus:
Thanks for the reference! Should this replacement be applied to all packages or to the dom package only? Since this issue has a label "scope:dom" I assume the later, correct?

staltz:
All packages where we use uglify. :)

maximilianmikus:
ok thanks, just wasn't sure ;)

maximilianmikus:
Ok, I think I got it set up, but how would you go about testing all these minified files? I think the test-suite does not cover them.

PR #632, opened 2017-06-09
perf(META): use Google Closure Compiler instead of uglify-js

maximilianmikus:
ISSUES CLOSED: #619

- [X] I ran `make test` for the all the packages

### DIFF
--- a/.scripts/minify.js
+++ b/.scripts/minify.js
@@ -1,0 +1,17 @@
+if (process.argv.length <= 3) {
+    console.log("Usage: " + __filename + " inputfile outputfile");
+    process.exit(-1);
+}
+var inputfile = process.argv[2];
+var outputfile = process.argv[3];
+var compiler = require('google-closure-compiler-js').compile;
+var fs = require('fs');
+var source = fs.readFileSync(inputfile, 'utf8');
+
+var compilerFlags = {
+  jsCode: [{src: source}],
+  languageIn: 'ES5'
+};
+
+var output = compiler(compilerFlags);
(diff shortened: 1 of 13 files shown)


maximilianmikus:
```
run/src/index.ts[1, 1]: This file has 316 lines, which exceeds the maximum of 300 lines allowed. Consider breaking this file up into smaller parts
```
I did not change run/src/index.ts ... From what I can figure, this is the reason the tests failed. Have not found any other error in the travis-ci logs.

staltz:
Don't worry about the unrelated errors. Looks good to merge. Thanks @maximilianmikus !

kylecordes:
@staltz I noticed the change to use Closure Compiler for better minification - great idea! It uses the JS version:

https://github.com/google/closure-compiler-js

which says: "This is an experimental release - meaning some features are not available and performance may not be on-par with the Java implementation."

I've been working with Closure Compiler over in other contexts (Angular, ClojureScript) and can confirm this description is true - the Java version is long-proven and intended for production use, the JS version is experimental in-progress. It might turn out to be sufficient already for Cycle's purpose, but warrants close attention as a member of the "experimental Closure Compiler JS" community, rather than the "using a long proven tool, don't need to pay any attention" community.

staltz:
@kylecordes this should be a rather harmless change because GCC is only used to produce a `dist/foo.min.js` file and these are usually just used for those cases where you directly import this script into your HTML. Usually that's not the case with serious frontend projects. There should be a bundler like Browserify or Webpack and then one of the post-processing steps should be minification. This PR does not affect that use case.

Issue #635, opened 2017-06-19
Bring @cycle/time to v1.0


staltz:
Continuing from #578, there are a couple things to be done before cycle/time is official.

- [ ] https://github.com/cyclejs/time/issues/37
- [ ] ?
- [ ] Make time-related xstream operators become extra, in xstream (as a PR, and release candidate)
- [ ] Migrate cycle/time README.md to `docs/content/api` (as a PR)
- [ ] Update all official examples (also in the site) to avoid xstream time-related operators (as a PR)
- [ ] Update documentation/tutorial if it mentions time anywhere
- [ ] Merge those PRs mentioned above and release v1.0

jvanbruegge:
I would add https://github.com/cyclejs/time/issues/29, because this will bring most likely some breaking changes for the architecture

Widdershin:
Also including https://github.com/cyclejs/time/issues/31 as it is important to nail down first.

Widdershin:
Opened #671 to implement extras.

@jvanbruegge I think we can do cyclejs/time#29 without breaking changes to the rest of the API, so it's my preference that we defer it until after v1.

I definitely still want to do it, but I want more time to discuss and consider the implementation without holding up the v1 release.

IssuehuntBot:
@issuehuntfest has funded $60.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/26316966/issues/635)

PR #636, opened 2017-06-20
feat(time): use setInterval to run events in background

Widdershin:
Previously, all events were driven by requestAnimationFrame. When a tab loses focus, requestAnimationFrame stops executing. By using setImmediate instead, events will fire in the background.

.animationFrames() and .throttleAnimation are still powered by requestAnimationFrame, so will not fire when the tab has lost focus.

Fixes https://github.com/cyclejs/time/issues/37. Progresses #635.

### DIFF
--- a/time/src/time-driver.ts
+++ b/time/src/time-driver.ts
@@ -9,6 +9,8 @@ import {makeThrottleAnimation} from './throttle-animation';
 import {runVirtually} from './run-virtually';
 import {TimeSource} from './time-source';
 const requestAnimationFrame = require('raf');
+const now = require('performance-now');
+require('setimmediate');
 
 function popAll(array: Array<any>): Array<any> {
   const poppedItems = [];
@@ -33,21 +35,34 @@ function runRealtime(
     paused = false;
   };
 
-  function processEvent(eventTime: number) {
+  function processFrameCallbacks(time: number) {
     if (paused) {
-      requestAnimationFrame(processEvent);
+      requestAnimationFrame(processFrameCallbacks);
       return;
     }
 
-    const time = eventTime;
     setTime(time);
 
     const currentCallbacks = popAll(frameCallbacks);
 
     currentCallbacks.forEach(callback => callback(time));
 
+    requestAnimationFrame(processFrameCallbacks);
+  }
+
+  requestAnimationFrame(processFrameCallbacks);
+
+  function processEvent() {
+    if (paused) {
+      setImmediate(processEvent);
+      return;
+    }
+
+    const time = now();
+    setTime(time);
+
     if (scheduler.isEmpty()) {
-      requestAnimationFrame(processEvent);
+      setImmediate(processEvent);
 
       return;
     }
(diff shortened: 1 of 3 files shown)


Widdershin:
@stevenmathews can you check this fixes your issue?

time/src/time-driver.ts
```diff
@@ -9,6 +9,8 @@ import {makeThrottleAnimation} from './throttle-animation';
 import {runVirtually} from './run-virtually';
 import {TimeSource} from './time-source';
 const requestAnimationFrame = require('raf');
+const now = require('performance-now');
```
jvanbruegge: `perfomance-now` is based on `process.hrtime`, so I would only use it in the mock time source (the one that will run most likely in node)

time/src/time-driver.ts
```diff
@@ -33,21 +35,34 @@ function runRealtime(
...
+
+  requestAnimationFrame(processFrameCallbacks);
+
+  function processEvent() {
+    if (paused) {
+      setImmediate(processEvent);
+      return;
+    }
+
+    const time = now();
```
jvanbruegge: As above I would use `performance.now()` here

time/src/time-driver.ts
jvanbruegge: I'm not even sure that this would work in the browser

Widdershin: I am using a module called `performance-now`, it defaults to using the `performance` api if available. This does work in the browser.

jvanbruegge: Should have looked at the source and not only the readme :S

staltz:
Merge with "rebase and merge" once you think it's ready. :)

Widdershin:
Okay so thanks to the testing done by @stevenmathews we determined that @cycle/time periodic had a bug where it would fall behind real time if left running in a background tab.

The easiest way to see this is to just combine an xs.periodic with a Time.periodic and see the difference.

This bug is because @cycle/time was not accounting for the possibility of event occurring slower than the given period. If we had a .periodic(1000), and the first event took 1200ms to execute because it was in a background tab, the next occurrence would be scheduled for 22000ms.

This is now fixed, and the second event will be scheduled for 2000ms regardless of when the first event occurs.

Performing the same test as above, combining an xs.periodic with Time.periodic shows that xstream is actually losing time compared to @cycle/time. This is because xs.periodic is powered by `setInterval`, which is not actually a reliable method of keeping exact time, as it can fluctuate in the background. You can verify that @cycle/time is correct by calling `performance.now()` in the console.

@jvanbruegge could you please check my last two commits 😄 I really appreciate you reviewing my code.

@stevenmathews could you confirm the issue you were experiencing is fixed with the latest commit?

staltz:
Good decision, and if anyone had asked me what to use to track exact time, I would *not* recommend xs.periodic. That's what I recommended in https://github.com/cyclejs/time/issues/37

Widdershin:
Thanks for the review and testing everyone, I'm happy to merge this.

PR #639, opened 2017-06-24
fix(dom): add support for "time" hyperscript-helper

ntilwalli:
- [X] I ran `make test FOO` for the package FOO I'm modifying


staltz:
Thanks @ntilwalli !

PR #642, opened 2017-07-03
Update model-view-intent.html

donaldpipowitch:
Looks like a copy and paste failure.

### DIFF
--- a/docs/model-view-intent.html
+++ b/docs/model-view-intent.html
@@ -283,7 +283,7 @@ Now, `main` is much smaller. But is it doing *one thing*? We still have `changeW
    const vdom$ = view(state$);
 
    return {
-     DOM: view(state$)
+     DOM: vdom$
    };
  }
 


Widdershin:
Good catch! Thanks for taking the time to fix this issue @donaldpipowitch, it's much appreciated 😄 🎉

donaldpipowitch:
Thanks for reviewing ❤️ 😄

staltz:
👍 for catching it. Just reminding, though: we need to update also the corresponding `.md` file otherwise these changes will be overriden.

Widdershin:
I thought about that after I merged it. I'll do that now 😄

staltz:
Maybe you could add a README.md under `/docs` so that in Github it shows, and we can add a disclaimer to edit the files under `/docs/content`, not the html.

Widdershin:
Good idea, #644 👍

donaldpipowitch:
Aaaah, good to know. I just followed the URL of the docs (https://cycle.js.org/model-view-intent.html) and looked for a corresponding file in `docs/`. I _think_ I wouldn't have read a warning like this one: https://github.com/cyclejs/cyclejs/pull/644/commits/4eca7d0f1fd0aad74a4ae5cb82c600e573deacc0.

But I _do_ look for top level comments very often to see, if a file was auto-generated. Maybe we can add a comment as the f[irst line after `<head/>`](https://github.com/cyclejs/cyclejs/blob/master/docs/model-view-intent.html#L4) or something like that in the generated HTML files?

staltz:
Or we could put a button "edit this file" in the html template, like we had in the old site.

donaldpipowitch:
That would be even better! I looked for a link like this before I switched to GitHub. (Maybe the comment could be generated additionally?)

PR #643, opened 2017-07-03
Add a simple routing example using @cycle/history

FeliciousX:
this is part of #583


staltz:
Thanks @FeliciousX for this PR!
One question before I merge: is it normal that fields Search, Hash, State are empty?

screen shot 2017-07-12 at 18 21 57

FeliciousX:
yeah. by default
- search is an empty string
- hash is an empty string
- state is undefined?

I am updating it with a better view of what's happening

staltz:
Thanks @FeliciousX! 
I'll merge this as a squash into one commit to add the example, then I'll refactor to bring the coding style more consistent with the other examples.

donaldpipowitch:
Cool. Thank you for the example 👍

PR #644, opened 2017-07-04
docs(META): add a docs README with a warning about editing html output

Widdershin:


staltz:
Thanks @Widdershin

PR #646, opened 2017-07-05
refactor(docs): add example with null scope

jvanbruegge:
when grepping the site for a `null` example, I overread the sentence three times.
An example is helpful here


staltz:
Looks good, thanks @jvanbruegge. Reading this again made me feel like clarifying the whole usage of isolate a bit better. Maybe I'll try that later

Issue #640, opened 2017-06-27
Update TypeScript to v2.4


ntilwalli:
I see this error when building with v2.4.1, it works fine with 2.3.4
```
TypeScript error: node_modules/@cycle/dom/lib/MainDOMSource.d.ts(6,22): Error TS2420: Class 'MainDOMSource' incorrectly implements interface 'DOMSource'. 
Types of property 'select' are incompatible.
Type '(selector: string) => DOMSource' is not assignable to type '<S extends DOMSource>(selector: string) => S'.
Type 'DOMSource' is not assignable to type 'S'.
```

oriSomething:
There are some changes the way interfaces works in TS ≥ 2.4.
Probably means that [this line](https://github.com/cyclejs/cyclejs/blob/master/dom/src/DOMSource.ts#L8) needs to be updated

staltz:
TypeScript v2.4 is a breaking change, so I'll rename this issue as an enhancement, it's not Cycle's fault that compilation fails.

whitecolor:
Will @ntilwalli  `skipLibCheck` compiler option help?

ntilwalli:
@whitecolor Thanks.  I'm back on 2.3.4 but I'll try that out when I get a chance.

PR #648, opened 2017-07-17
Update all to TypeScript 2.4

staltz:
For issue #640


staltz:
Could either @jvanbruegge or @Widdershin take a quick look at this?

most-run/src/index.ts
```diff
@@ -51,8 +51,8 @@ export function run<So extends Sources, Si extends Sinks>(
   main: (sources: So) => Si,
   drivers: Drivers<So, Si>,
 ): DisposeFunction {
-  const {run} = coreSetup(main, drivers);
-  return run();
+  const program = coreSetup(main, drivers);
+  return program.run();
```
jvanbruegge: Why this change?

staltz: Because we have a new lint rule that warns against shadowing var names. In this case, the local var `run` was conflicting with the module-level exported `run` function.

staltz:
Cool, I'll merge it

PR #650, opened 2017-07-20
[WIP] feat(http): adding a skeleton for a new mock http source helper

### LINKED ISSUE #567: Add mockHTTPSource helper

Widdershin:
Carrying on from #571, had to make a new pull request because I couldn't push to the old branch.

Fixes #567

### DIFF
--- a/http/test/node/mock-http-source.ts
+++ b/http/test/node/mock-http-source.ts
@@ -1,0 +1,9 @@
+import 'mocha';
+import * as assert from 'assert';
+import {mockHTTPSource} from '../../lib/index';
+
+describe('mockHttpSource', function () {
+  it('should be in accessible in the API', function () {
+    assert.strictEqual(typeof mockHTTPSource, 'function');
+  });
+});
--- a/http/test/node/mock-http-source.ts
+++ b/http/test/node/mock-http-source.ts
@@ -1,9 +1,21 @@
 import 'mocha';
 import * as assert from 'assert';
-import {mockHTTPSource} from '../../lib/index';
+import { mockHTTPSource } from '../../lib/index';
+import xs from 'xstream';
 
 describe('mockHttpSource', function () {
   it('should be in accessible in the API', function () {
     assert.strictEqual(typeof mockHTTPSource, 'function');
+  });
+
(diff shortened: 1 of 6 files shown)


http/test/node/mock-http-source.ts
```diff
@@ -1,63 +1,23 @@
...
-      complete: () => done(),
-    });
-  });
-
-  it('supports categorized streams', (done) => {
-    const response: any = {
-      text: 'Hello world',
+import {setup} from '../../../run/src';
+
+describe.only('mockHTTPDriver', function() {
```
staltz: I suppose `describe.only` is temporary? :)

Widdershin: Most definitely. This latest commit doesn't even compile yet 😸 

I need a way to turn a `Partial<Response>` into something that implement the `Response` interface. Ideally I want to match superagent's behaviour perfectly, but most of the logic I need to utilize is tightly coupled with performing side effects.

In the past I've used https://github.com/M6Web/superagent-mock, so hopefully I can reuse some of the work they've done.

grushetsky:
@Widdershin, do you need any help on this one? We'd like to use mocked HTTP source too!

Widdershin:
@grushetsky I'd love some help, been meaning to finish this off for a while but I've been a bit swamped.

The next main thing that needs to be done is finishing the implementation of `partialResponseToMockResponse` (https://github.com/cyclejs/cyclejs/pull/650/files#diff-f0a89f91a42b58374b1c8cec24cd397cR59). This effectively needs to replicate the behaviour of the superagent internals. I was going to looking into extracting part of [superagent-mock](https://github.com/M6Web/superagent-mock), but I have yet to do that.

Hasty explanation so let me know if you want more detail, or if that's enough to go on.

jvanbruegge:
Made irrelevant due to the rewrite

PR #653, opened 2017-07-27
Put svg attributes inside attrs object

johnrees:
I don't know why the CHANGELOG has been affected by this commit, I just edited the appropriate files and ran `make test dom` & `make commit`

related to https://gitter.im/cyclejs/cyclejs?at=5979ccdac101bc4e3ac5b82b

### DIFF
--- a/dom/test/browser/src/isolation.ts
+++ b/dom/test/browser/src/isolation.ts
@@ -626,7 +626,7 @@ describe('isolation', function() {
     function App(sources: {DOM: MainDOMSource}) {
       const triangleElement$ = sources.DOM.select('.triangle').elements();
 
-      const svgTriangle = svg({width: 150, height: 150}, [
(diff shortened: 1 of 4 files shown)


dom/test/browser/src/isolation.ts
```diff
@@ -626,7 +626,7 @@ describe('isolation', function() {
     function App(sources: {DOM: MainDOMSource}) {
       const triangleElement$ = sources.DOM.select('.triangle').elements();
 
-      const svgTriangle = svg({width: 150, height: 150}, [
```
jvanbruegge: @stalz Why did thus test not fail?

staltz: It could be that the width and height are not important for the test success/failure. I think it only affects how big the SVG looks like, and therefore it's just a visual thing.

jvanbruegge:
Thanks for the PR!

Issue #592, opened 2017-04-27
HTTP driver erroneously replays previous request


ntilwalli:
While updating to Cycle Unified I noticed that some of my HTTP requests from a previous component are being replayed when a new component gets swapped in.  I traced the issue to `makeSinkProxies` which is improperly defaulting all proxies to `MemoryStream`. Not all sinks should be represented using MemoryStream, the HTTP stream being a clear example.

https://github.com/cyclejs/cyclejs/blob/69f932ca01abafad8b08d111c0d639cb9b013f8b/run/src/index.ts#L83

```js
function makeSinkProxies<So extends Sources, Si extends Sinks>(
                        drivers: Drivers<So, Si>): SinkProxies<Si> {
  const sinkProxies: SinkProxies<Si> = {} as SinkProxies<Si>;
  for (const name in drivers) {
    if (drivers.hasOwnProperty(name)) {
      sinkProxies[name] = xs.createWithMemory<any>();  // <-------- THIS IS WRONG FOR HTTP
    }
  }
  return sinkProxies;
}
```

staltz:
Okay, this is a nasty problem.

I remember reverting to normal Stream, then other issues started popping up, and we used MemoryStream, I even left a comment behind so we would remember exactly why is it a memory stream: https://github.com/cyclejs/cyclejs/blob/69f932ca01abafad8b08d111c0d639cb9b013f8b/run/src/index.ts#L44-L58

```js
/**
 * Sink proxies should be MemoryStreams in order to fix race conditions for
 * drivers that subscribe to sink proxies "later".
 *
 * Recall that there are two steps:
 * 1. Setup (sink proxies -> drivers -> sources -> main -> sink)
 * 2. Execution (also known as replication: sink proxies imitate sinks)
 *
 * If a driver does not synchronously/immediately subscribe to the sink proxy
 * in step (1), but instead does that later, if step (2) feeds a value from the
 * sink to the sink proxy, then when the driver subscribes to the sink proxy,
 * it should receive that value. This is why we need MemoryStreams, not just
 * Streams. Note: Cycle DOM driver is an example of such case, since it waits
 * for 'readystatechange'.
 */
```

I'm not sure what would be the solution. Maybe `UglyFixStream extends Stream` where UglyFixStream replays the remembered value just once, then from that moment onwards behaves like a Stream.

ntilwalli:
Replaying only the first emission will not fix the problem. 

[...]

Issue #222, opened 2016-02-25
makeDOMDriver does not handle being called before DOM Ready


staltz:
_From @WorldMaker on October 20, 2015 16:19_

I created a test bundle of all my scripts (using `jspm bundle-sfx` if you are curious, but that shouldn't matter) and when attempting to run as a bundle I hit the case where the browser was running the `makeDOMDriver` function prior to the DOM Ready. (Interestingly the DOM in this case was just the single DIV I'm currently using as a container, so kind of amazing the browsers manage to run the scripts before evaluating the DIV...)

Does it make sense for `makeDOMDriver` to handle the case where it is being called before DOM Ready?

_Copied from original issue: cyclejs/cycle-dom#56_

staltz:
Yeah I'll mark this as a bug because makeDOMDriver should indeed start doing its stuff after the DOM is ready, but nothing will change on the API side. You will still call makeDOMDriver whenever you want.

chadrien:
I'll be working on this.

This is simply about adding `document.addEventListener(`DOMContentLoaded`, () => {/**/})` at the right place in `makeDOMDriver` right?

staltz:
@chadrien I think we should wait for a major rewrite to Cycle DOM before doing this. In other words, let's pause any development on Cycle DOM master, unless it's a critical bug. Check out this branch https://github.com/cyclejs/dom/tree/diversity where we are rewriting Cycle DOM in TypeScript and Snabbdom. It'll be a breaking change because virtual-dom API differs from snabbdom just a tiny bit.

I'll remove `help wanted` from this issue for now. I'll check other issues to see if `help wanted` is for real.

staltz:
See also comments of duplicate https://github.com/cyclejs/core/issues/301

Cmdv:
@staltz got a fix but TypeScript doesn't like `Property 'readystate' does not exist on type 'EventTarget'`.

might be a fix [here](http://stackoverflow.com/questions/35789498/new-typescript-1-8-4-build-error-build-property-result-does-not-exist-on-t) but then I'm not sure how I can make a test that would be reasonable!! basically I've been no help!! 

But here is the [PR](https://github.com/cyclejs/dom/pull/124/files)

artfuldev:
if this is still a bug I can work on this...

Cmdv:
Take my PR as a reference if you like @artfuldev 😄 
I've not bothered learning TypeScript as I went towards PureScript instead! So this bug is still open

[...]

PR #658, opened 2017-07-31
Driver buffers to solve race conditions from issue #592

staltz:
Fixes all 3 `run`, `dom`, `http`. There is the possibility other community drivers may take the same buffer approach used in Cycle DOM and Cycle HTTP.

### DIFF
--- a/docs/content/documentation/drivers.md
+++ b/docs/content/documentation/drivers.md
@@ -276,10 +276,92 @@ run(main, {
 
 Notice we have the `peerId` specified when the driver is created in `makeSockDriver(peerId)`. If the `main()` needs to dynamically connect to different peers according to some logic, then we shouldn't use this API anymore. Instead, we need the driver function to take instructions as input, such as "connect to peerId", or "send message to peerId". This is one example of the considerations you should take when designing a driver API.
 
+## Resistance to race conditions
+
+Complex drivers such as the official DOM and HTTP drivers should be able to handle all sorts of corner cases, such as synchronous race conditions or ordering of sinks in the sinks objects. This is because subscriptions ("adding a listener") happen only in drivers, and there are different types of subscription bugs that may surface only in driver code.
+
+These corner cases are very rare, but possible, such as issue [#592](https://github.com/cyclejs/cyclejs/issues/592). Fortunately, with a simple technique in driver code, you can defend your code against these possibly rare bugs. We could have put a **generic solution** for this in Cycle Run, but it would have a small negative effect to all Cycle.js apps, so we decided to put **specific solutions** in each driver where this may occur.
+
+If your driver is similar to the DOM driver, it meets these conditions:
+
+- The *source* depends on the `foo$` which depends on the *sink*
+- `foo$` is hot (multicasted), like most xstream Streams are
+- `foo$` is subscribed with a dummy listener, even if the source is not used in the `main`
+- `foo$` may emit synchronously when subscribed
+
+Those conditions may lead to the following rare race condition:
+
+1. `foo$` is subscribed
+2. `foo$` emits synchronously, but the value is ignored by the dummy listener
+3. The *source* is used by a chain of streams and eventually subscribed
+4. The *source* does not get the value dropped by step 2
+
+To solve this case, you can use a simple buffer array technique. Suppose your driver looks like this:
+
+```js
+function myDriver(sink) {
+  const foo$ = sink.map(doSomeProcessing);
+
+  // Dummy listener to make sure the processing happens
+  foo$.addListener({
+    next: () => {},
+    error: () => {},
+    complete: () => {},
+  })
+
+  const source = foo$.map(prepareTheSource);
+
+  return adapt(source);
+}
+```
+
+Then you can:
+
+- Add a `buffer` array and a boolean `isBufferOpen`
+- Add a stream `tooEarlyFoo$` that consumes the buffer
+- In the dummy listener for `foo$`, push values to the `buffer`
+- Source to depend on both `tooEarlyFoo$` and `foo$`
+
+Such as the following solution:
+
+```diff
+ function myDriver(sink) {
++  let isBufferOpen = true;
++  const buffer = [];
++  const tooEarlyFoo$ = xs.create({
++    start(listener) {
++      while (buffer.length > 0) {
++        listener.next(buffer.shift());
++      }
++      isBufferOpen = false;
++    },
++    stop() {},
++  });
+   const foo$ = sink.map(doSomeProcessing);
+
+   // Dummy listener to make sure the processing happens
+   foo$.addListener({
+-    next: () => {},
++    next: (x) => {
++      if (isBufferOpen) buffer.push(x);
++    },
+     error: () => {},
+     complete: () => {},
+   })
+
+-  const source = foo$.map(prepareTheSource);
++  const source = xs.merge(foo$, tooEarlyFoo$).map(prepareTheSource);
+
+   return adapt(source);
+ }
+```
+
+These additions are enough to make your driver resistant to race conditions in various use cases.
+
 ## Extensibility
 
-Cycle *Core* is a very small framework, and Cycle *DOM*'s Driver is available as an optional plugin for your app. This means it is simple to replace the DOM Driver with any other driver function providing interaction with the user.
+Cycle.js is a very small framework, and Cycle *DOM*'s Driver is available as an optional plugin for your app. This means it is simple to replace the DOM Driver with any other driver function providing interaction with the user.
 
 You can for instance fork the DOM Driver, adapt it to your preferences, and use it in a Cycle.js app. You can create a driver to interface with sockets. Drivers to perform network requests. Drivers meant for Node.js. Drivers that target other UI trees, such as `<canvas>` or even native mobile UI.
 
-As a framework, it cannot be compared to monoliths which have ruled web development in the recent years. Cycle.js itself is after all just a small tool and a convention to create reactive dialogues with the external world using reactive streams.
+As a framework, it cannot be compared to monoliths in web development in the recent years. Cycle.js itself is after all just a small tool and a convention to create reactive dialogues with the external world using reactive streams.
--- a/docs/content/documentation/drivers.md
+++ b/docs/content/documentation/drivers.md
@@ -360,8 +360,8 @@ These additions are enough to make your driver resistant to race conditions in v
 
 ## Extensibility
 
-Cycle.js is a very small framework, and Cycle *DOM*'s Driver is available as an optional plugin for your app. This means it is simple to replace the DOM Driver with any other driver function providing interaction with the user.
+Cycle.js drivers, such as the DOM Driver, are available as optional plugins for your app. This means it is simple to replace a driver with another one that has the same expected sinks and sources.
 
-You can for instance fork the DOM Driver, adapt it to your preferences, and use it in a Cycle.js app. You can create a driver to interface with sockets. Drivers to perform network requests. Drivers meant for Node.js. Drivers that target other UI trees, such as `<canvas>` or even native mobile UI.
+For instance, you can fork the DOM Driver, adapt it to your preferences, and use it in a Cycle.js app. You can create a driver to interface with sockets. Drivers to perform network requests. Drivers meant for Node.js. Drivers that target other UI trees, such as `<canvas>` or even native mobile UI.
 
-As a framework, it cannot be compared to monoliths in web development in the recent years. Cycle.js itself is after all just a small tool and a convention to create reactive dialogues with the external world using reactive streams.
+As a framework, it cannot be compared to monoliths in web development in the recent years. Cycle.js itself is after all just a small tool and a convention to create reactive dialogues with the external world using streams.
--- a/dom/package.json
+++ b/dom/package.json
@@ -34,7 +34,7 @@
   "typings": "lib/index.d.ts",
   "types": "lib/index.d.ts",
   "dependencies": {
-    "@cycle/run": "*",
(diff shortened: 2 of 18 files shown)


staltz:
The tests don't pass in CI because the DOM/HTTP driver needs the run from this PR too. Maybe I should make a `rc`.

jvanbruegge:
I think this would be good

staltz:
Good point, I can write that in release notes

jvanbruegge:
I would add it as paragraph in the "writing a driver" doc. In the release notes it will drown

staltz:
That's also a good point 👍

staltz:
@jvanbruegge can you check the drivers.md change I made?

I'll make `rc` versions next.

Widdershin:
Nice documentation 👍 💯

whitecolor:
@staltz 
I think there is actually no need in the `tooEarly` buffer in HTTP driver. Requests will get to the driver only after sources in the Main are subscribed, so it has all the chances not to miss any response. **So I don't see the reason to handle delayed subscription inside Main.** The HTTP response$$ source should be hot and not remembered. If you think otherwise I really would like to hear a reason why the first delayed subscription should be handled using buffers, what are the real case and problems will it solve.

whitecolor:
Also a couple of notices about DOM driver:

1) There is a need in only last early element which represents the state of vtree, so no need to gather a buffer. Probably better handle too early `vnode$` not `rootElement$`, and it should take the last element from vnode$ prior ready to patch. I believe this can be just done by just converting it to MemoryStream.
3) I see from the code that issue https://github.com/cyclejs/cyclejs/issues/222 is still not being handled correctly: it is trying to [get container element](https://github.com/cyclejs/cyclejs/blob/51834dca6b1a07feb253d3b5b88748cb4698df6f/dom/src/makeDOMDriver.ts#L63) when content may not be still ready.

staltz:
@whitecolor Thanks for taking a careful look into this.

About the HTTP case:

> Requests will get to the driver only after sources in the Main are subscribed

What makes you think this is true? I think there are cases where request sinks arrive to the driver regardless of sources being subscribed or not, such as fire-and-forget POST requests. If I'm not mistaken, those tests added to cycle/http wouldn't pass without the buffer.

> 1. There is a need in only last early element which represents the state of vtree, so no need to gather a buffer. Probably better handle too early vnode$ not rootElement$, and it should take the last element from vnode$ prior ready to patch. I believe this can be just done by just converting it to MemoryStream.

Good point, I'll try that.

> 2. I see from the code that issue #222 is still not being handled correctly: it is trying to get container element when content may not be still ready.

That's very true, I'll do something about it in this branch.

whitecolor:
> Requests will get to the driver only after sources in the Main are subscribed

> What makes you think this is true? 

Because of replication buffers inside run all sinks will be subscribed before any request will get to HTTP driver sink and respectively response$ to driver source, so all subscription to sources that are made without any delay will be activated.

> I think there are cases where request sinks arrive to the driver regardless of sources being subscribed or not,

This is true but only if the user has some delayed subscription to resonse$$ source, that is activated after requests is send to sink - this just should be the responsibility of the user, otherwise, there is possible inconsistent behaviour, see these two cases:

First case when we have an immediate subscription to HTTP source that will never miss response and also we have delayed subscription.

```ts
const Main({HTTP}) => {
  // this subscribtion will anyway get any response$, never miss very early response$
  HTTP.select().. 

  // next delayed subscription will miss responses that come after 1 second
  // it will NOT get any early responses from HTTP buffer
  // because response$$ was already subscribed
  of('').compose(delay(1000))
    .map(_ => HTTP.select()) 
}
```

But in next the case we don't have a first immediate subscription and due to introduced buffers in HTTP, delayed subscription will get early responses from the buffer.
```ts
const Main({HTTP}) => {
  // such subscription will get all responses including early responses from buffer
 // so for first subscriber response$$ works as cold source
 // it behaves differently comparing to the prior case with two subscribtions
 // that is why I say it is inconsistent behavior
  of('').compose(delay(1000))
    .map(_ => HTTP.select()) 
}
```

staltz:
I fixed some issues related to the DOM driver that @whitecolor mentioned.

About the HTTP driver, however, I still can't see this happening:

> Requests will get to the driver only after sources in the Main are subscribed
> ...
> Because of replication buffers inside run all sinks will be subscribed before any request will get to HTTP driver sink and respectively response$ to driver source, so all subscription to sources that are made without any delay will be activated.

Because chained subscriptions happen *depth-first*, not *breadth-first*, and in many cases a sink A depends on a source B, and source B depends on sink B. So it's not true that "sink data will get to the driver after sources in Main are subscribed".

Example:

```js
    function child(sources: any, num: any) {
      const vdom$ = sources.HTTP
        .select('cat')
        .mergeAll()
        .map((res: any) => 'My name is ' + res.text);

      const request$ = num === 1
        ? Rx.Observable.of({
            category: 'cat',
            url: uri + '/hello',
          })
        : Rx.Observable.never();

      return {
        HTTP: request$,
        DOM: vdom$,
      };
    }
```

Cycle run will subscribe to HTTP sink, which subscribe to request$, and the data will synchronous go down the chain. Then, synchronously, Cycle run would subscribe to the DOM sink, which subscribes to HTTP source.

For that reason, if I just remove the new buffer technique from Cycle HTTP, tests fail.

The other HTTP bug, about delayed subscriptions, seems real and I'm working on it now.

staltz:
All tests pass now. Can someone review this PR? All good to publish, or not?

docs/content/documentation/drivers.md
```diff
@@ -276,10 +276,92 @@ run(main, {
...
+   return adapt(source);
+ }
+```
+
+These additions are enough to make your driver resistant to race conditions in various use cases.
+
 ## Extensibility
 
-Cycle *Core* is a very small framework, and Cycle *DOM*'s Driver is available as an optional plugin for your app. This means it is simple to replace the DOM Driver with any other driver function providing interaction with the user.
+Cycle.js is a very small framework, and Cycle *DOM*'s Driver is available as an optional plugin for your app. This means it is simple to replace the DOM Driver with any other driver function providing interaction with the user.
```
jvanbruegge: I would generalize this beyond DOM

docs/content/documentation/drivers.md
```diff
@@ -360,7 +360,7 @@ These additions are enough to make your driver resistant to race conditions in v
 
 ## Extensibility
 
-Cycle.js is a very small framework, and Cycle *DOM*'s Driver is available as an optional plugin for your app. This means it is simple to replace the DOM Driver with any other driver function providing interaction with the user.
+Cycle.js is a very small framework, and drivers, such as the DOM Driver, are available as optional plugins for your app. This means it is simple to replace a driver with another one that has the same expected sinks and sources.
```
jvanbruegge: Maybe: `Cycle.js is a very small framework, at its core it is just glue for drivers, such as [...] which are available [...].`?
Sounds a bit off ATM

whitecolor:
> Cycle run will subscribe to HTTP sink, which subscribe to request$, and the data will synchronous go down the chain. Then, synchronously, Cycle run would subscribe to the DOM sink, which subscribes to HTTP source.

I didn't check this case (code) in practice, but:

1) Cycle subscribes to HTTP sink (request) - requests are gathered in Cycle inner HTTP buffer, till all sinks are subscribed.
2) Cycle subscribe to DOM sink (subscribes to HTTP source)
3) After sinks are subscribed, requests from inner Cycle HTTP buffer go to driver and responses go to HTTP source (which is subscribed in step 2).

Is the order theoretically correct? Why there is buffer inside HTTP driver needed?

staltz:
@whitecolor your theory is correct regarding how cycle/run buffers initial sink events until all sinks are subscribed. However, this issue is about nested child components which are dynamically called, potentially much after cycle/run replicateMany did its job.

This is really tricky to explain, so I ran [this test](https://github.com/cyclejs/cyclejs/blob/41d42dfd804958251196a16d98515648319e83e9/run/test/index.ts#L426) first **without driver buffers** and with comments in key places:

```
HTTP driver will subscribe to HTTP source
subscribe to HTTP source
cycle/run replicate sink to sink proxies
cycle/run replication is forwarding DOM sink to sink proxy
child num is 1
subscribe to child HTTP sink
cycle/run replication is forwarding HTTP sink to sink proxy
subscribe to child DOM sink
cycle/run replication is forwarding DOM sink to sink proxy
```

Then **with** driver buffer:

```
cycle/run calls all drivers with sink proxies
HTTP driver will subscribe to HTTP source
subscribe to HTTP source
cycle/run replicate sink to sink proxies
cycle/run replication is forwarding DOM sink to sink proxy
child num is 1
subscribe to child HTTP sink
cycle/run replication is forwarding HTTP sink to sink proxy
HTTP sink (request) is being transformed into an HTTP source (response)
HTTP driver is putting source event  { body: { name: 'Louis' } }  in the driver buffer
subscribe to child DOM sink
subscribe to HTTP driver earlySource
HTTP driver is consuming source event  { body: { name: 'Louis' } }  from the driver buffer
HTTP source is emitting
cycle/run replication is forwarding DOM sink to sink proxy
cycle/run replication is forwarding DOM sink to sink proxy
child num is 3
subscribe to child HTTP sink
subscribe to child DOM sink
cycle/run replication is forwarding DOM sink to sink proxy
child num is 5
subscribe to child HTTP sink
subscribe to child DOM sink
cycle/run calls all drivers with sink proxies
HTTP driver will subscribe to HTTP source
subscribe to HTTP source
cycle/run replicate sink to sink proxies
cycle/run replication is forwarding DOM sink to sink proxy
child num is 1
subscribe to child DOM sink
subscribe to HTTP driver earlySource
subscribe to child HTTP sink
cycle/run replication is forwarding HTTP sink to sink proxy
HTTP sink (request) is being transformed into an HTTP source (response)
HTTP source is emitting
cycle/run replication is forwarding DOM sink to sink proxy
cycle/run replication is forwarding DOM sink to sink proxy
child num is 3
subscribe to child DOM sink
subscribe to child HTTP sink
cycle/run replication is forwarding DOM sink to sink proxy
child num is 5
subscribe to child DOM sink
subscribe to child HTTP sink
```

The reason why this occurs is that these processes happen in this order:

1. In cycle/run, all the drivers are called, with sink proxies instead of real sinks
2. In one of the drivers, the `source` will be created as a simple stream definition based on the given `sink`. See diagram below
3. In that driver, the `source` will be subscribed to, this is the red line in the diagram. The listener will forward data to the driver buffer
4. In cycle/run the `main` will be called with those sources
5. According to this test example, the HTTP source will not be subscribed **yet** because the `child` component has not yet been called
6. In cycle/run, `replicateMany` happens, but all of that buffer logic is irrelevant since no HTTP sink events happened so far (because `child` has not been called)
7. After 1 second has passed, `child` will be called
8. The child's HTTP sink is subscribed, and this data arrives in the HTTP driver
9. The HTTP driver performs the real network request and produces a source event (the response)
10. The response data gets stored in the driver buffer (the red line in the diagram)
11. Still part of calling `child` function, now the DOM sink will be subscribed, which in turn will subscribe to `outputSource` from the HTTP driver
12. Finally, `earlySource` from the HTTP driver is subscribed, so the driver buffer is consumed
13. `outputSource` emits response data

photo_2017-09-19_13-27-21

staltz:
Describing the issue carefully above made me think we probably have another bug left unresolved.

The driver buffer is a one-time mechanism. It only works for the first `child` call that wants to consume `outputSource`. If we have a totally different `child2` component that will be dynamically called later in the same style that `child` was called, the driver buffer won't work anymore (it was already used) and it will be a bug.

I think the root cause is that we'd like to do **synchronous yet breadth-first** subscribes of (any) component's sinks.

staltz:
Yeah, I'm getting convinced we need a scheduler for this in xstream. Right now we're using the recursive scheduler which is also the default in RxJS v5:

```js
var a$ = Rx.Observable.of(1, 2)
  .merge(Rx.Observable.never())
  .share();
var b$ = a$.map(x => x * 10);

a$.subscribe(x => console.log('a' + x));
b$.subscribe(x => console.log('b' + x));
```

will output in the console

```
a1
a2
```

But the following code using the `asap` scheduler:

```js
var a$ = Rx.Observable.of(1, 2, Rx.Scheduler.asap)
  .merge(Rx.Observable.never())
  .share();
var b$ = a$.map(x => x * 10);

a$.subscribe(x => console.log('a' + x));
b$.subscribe(x => console.log('b' + x));
```

outputs

```
a1
b10
a2
b20
```

Unfortunately the `asap` scheduler has a downside: less performant than the recursive scheduler, and makes for less readable stack traces.

staltz:
Or perhaps just use most.js:

```js
var a$ = most.of(1).concat(most.of(2))
  .merge(most.never())
  .multicast();
var b$ = a$.map(x => x * 10);

a$.subscribe({next: x => console.log('a' + x)});
b$.subscribe({next: x => console.log('b' + x)});
```

```
a1
b10
a2
b20
```

:/

staltz:
Both RxJS `asap` scheduler and most.js use `setImmediate` or `Promise.resolve` to solve this. Which takes us back to [this comment](https://github.com/cyclejs/cyclejs/issues/592#issuecomment-315207550).

Honestly, I'm inclined to just implementing the solution Tylor suggested:

```js
const runOnNextTick = f => typeof Promise !== 'undefined'
  ? Promise.resolve(f).then(g => g())
  : setTimeout(f, 0)
```

Because in most browsers it would use Promise.resolve which is very fast and wouldn't incur any noticeable delay (only `setTimeout` would).

staltz:
Note: this PR is outdated because #699 is better, but let's keep this one around until #699 is merged.

geovanisouza92:
This seems ready to be closed.

PR #662, opened 2017-08-04
Add ES6 module builds

### LINKED ISSUE #660: @cycle/time: cjs bundle means lodash deps for mocking are pulled in on production apps
mock-time-source.ts uses package combine-errors, which uses package lodash.uniqby which then pulls in a bunch of extra lodash stuff which bloats the bundle considerably if you're not already using lodash.

lodash.uniqby brings in lodash.iteratee (56kb) among others - all in all it adds up to almost 100kb of extra code that won't be used:
extralodashdeps

This wouldn't be a problem if we were able to use ES module versions of the packages (rather than bundled cjs) and a treeshaking/dead code eliminating bundler as in theory it could remove the unused modules.

My solution is to split the mock vs non-mock timesource files so in your tests you can pull in the mock time source (including any dependencies needed, e.g. lodash) but your app can pull in the skinnier time driver alone.
Here's an example:
https://github.com/cyclejs/time/compare/master...willheslam:separate-test-deps?expand=1

**Expected behavior:**
No lodash deps in production bundle that does not use mock sources

**Actual behavior:**
lodash.iteratee and other deps are included even  when not using mock sources

cycle/time: ^0.8.0

### LINKED ISSUE #661: Provide ES6 bundles
You can specify a `module` entry in `package.json` that is used to provide ES6 modules besides commonjs. (More info [here](https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c)).

I would like to tackle that.
Will resolve #660

jvanbruegge:
fixes #661 
fixes #660

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ setup :
 		cd $$d ; yarn install ; cd .. ;\
 		echo "" ;\
 	done < .scripts/RELEASABLE_PACKAGES
-	@make lib all
+	@make lib
 
 prettier-all :
 	@echo "Formatting all source files using prettier"
@@ -76,7 +76,8 @@ lib :
 	else \
 		rm -rf $(ARG)/lib/ ;\
 		mkdir -p $(ARG)/lib ;\
-		$(TSC) --project $(ARG) ;\
+		$(TSC) --project $(ARG) --module commonjs --outDir $(ARG)/lib ;\
+		$(TSC) --project $(ARG) --module es6 --outDir $(ARG)/lib/es6 ;\
 		echo "✓ Compiled TypeScript to lib\n" ;\
 	fi
 
--- a/dom/package.json
+++ b/dom/package.json
@@ -30,6 +30,7 @@
     "virtual-dom"
   ],
   "main": "lib/index.js",
+  "module": "lib/es6/index.js",
   "typings": "lib/index.d.ts",
   "types": "lib/index.d.ts",
   "dependencies": {
--- a/dom/tsconfig.json
+++ b/dom/tsconfig.json
@@ -1,7 +1,6 @@
 {
   "extends": "../tsconfig.common.json",
   "compilerOptions": {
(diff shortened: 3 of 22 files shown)


jvanbruegge:
@staltz Should be ready to merge

staltz:
@jvanbruegge Good catch! Looks good to merge, except for the commit messages. I could do a squash if you don't mind. Then, we'll have to `make release-minor` for each package in order to get pkg.module making some difference.

jvanbruegge:
Yeah, sqash them :)

jvanbruegge:
I sqashed it myself.
Yeah, a new release for everything would be needed

staltz:
All packages got new minor versions 👍

jvanbruegge:
:+1:

5310:
I noticed that two of the source files under `@cycle/dom` use `require()`s to polyfill ES6 maps. Unfortunately, these are retained in the ES6 builds:
- https://github.com/cyclejs/cyclejs/blob/master/dom/src/IsolateModule.ts#L3
- https://github.com/cyclejs/cyclejs/blob/master/dom/src/makeDOMDriver.ts#L14

These break the library under Rollup (using `rollup-plugin-node-resolve` and bundling to ES6 modules themselves) when the ES6 build is imported as is and the browser throws an exception when encountering the unexpected `require()`s. 

Changing these require statements to `import MapPolyfill from 'es6-map'` seems to work out just fine. If this is an acceptable change, should I submit this as a PR?

jvanbruegge:
Yeah, do it

5310:
Should I leave the `es6-map` module type-less or at least add in a `declare module 'es6-map';` definition file? This still raises untyped function calls warnings, but I'm afraid I'll need help with providing proper definitions as I don't know much about TypeScript.

jvanbruegge:
Leave it like that, that's just a Shim for the ES6 map which definition is included in Typescript itself

PR #665, opened 2017-08-11
refactor(time): enhanced type definitions for MockTimeSource#diagram

jsamr:
diagram method has now three signatures
- the first with one argument, returning `Stream<number | string>`
- the second with two arguments, with the latest an `Array<T>` of values, returning
`Stream<T>`
- the third with two arguments, with the latest an `ObjectDictionary<T>` of values,
returning `Stream<T>`

I didn't add tests since this addition aims to reinforce an existing contract.

### DIFF
--- a/time/src/time-source.ts
+++ b/time/src/time-source.ts
@@ -4,6 +4,10 @@ import {Frame} from './animation-frames';
 export type Operator = <T>(stream: Stream<T>) => Stream<T>;
 export type Comparator = (actual: any, expected: any) => void;
 
+export interface ObjectDictionary<T> {
+  [key: string]: T;
+}
+
 export interface TimeSource {
   animationFrames(): Stream<Frame>;
   delay(delayTime: number): Operator;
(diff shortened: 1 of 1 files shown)


Widdershin:
Hey @sveinburne,

Thanks for this contribution! And for moving your pull request over.

Sorry for my delay in getting back to you.

Would you be able to take a look at the CI failure and see if you can determine the cause?

Widdershin:
Will merge this after the fix for #672 lands.

sveinburne:
Hi @Widdershin, sorry for my late answer ! Was in holiday. Will take a look this week-end, 
cheers

Widdershin:
@sveinburne no worries! 😄 @jvanbruegge and I are still sorting out our approach to #672 so we wouldn't have been able to merge this in any case.

sveinburne:
All right ! That is why I passed the tests locally but not on CI server I guess !

Issue #667, opened 2017-08-17
@cycle/dom elements lose their EventListener when moved in another part of the DOM


atomrc:
**Code to reproduce the issue:**
Here is a bin that reproduces the bug https://www.webpackbin.com/bins/-KrjXqknNMZskaEo273r
Step to reproduce:
- Button 1 and Button 2 should update state when clicked ;
- click the `sort` button ;
- Button 2 still works ;
- Button 1 doesn't react to clicks anymore.

**Expected behavior:**
The button moved to the bottom should update state

**Actual behavior:**
The button moved to the bottom doesn't do anything when clicked

**Versions of packages used:**
@cycle/dom v18.1.0
@cycle/run v3.2.0
cycle-onionify v4.0.0-rc-14

I think this bug has to do with the way the `EventDelegator` cleans vnodes when they are treated by snabbdom https://github.com/cyclejs/cyclejs/blob/b482e206b1425dda95c63ad1ca872189b5c0d93e/dom/src/IsolateModule.ts#L135. One fix could be to batch the removal of listener once snabbdom has completely finished its work?

Here is a gif that illustrate what happens 
![bug](https://s3-eu-west-1.amazonaws.com/atomrc-random/out.gif)

Widdershin:
Yes, we absolutely should fix this. Thanks for the reproducer.

If anyone is interested in making a pull request, please comment in here. 😄

atomrc:
I might be interested in patching this. 
The idea I have in mind is to: 
- whenever a node is removed, add its key to an array of vnodes to clean ;
- if a removed vnode is added latter on, just remove it from the array of vnodes to clean ;
- when snabbdom has finished its work, actually trigger the vnodes cleaning.

I am not really familiar with snabbdom, but it seems that the `post` hook seem perfect for that use case, isn't it? https://github.com/snabbdom/snabbdom#hooks

atomrc:
@Widdershin so I was able to write a test that reproduce the error. Can you give me some feedback on it? https://github.com/atomrc/cyclejs/commit/7e6a15a793201fad533e7913d650cb35b1c7dc77

Next step would be actually fixing it :)

Frikki:
Maybe you can draw some inspiration from [Motorcycle](https://github.com/motorcyclets/motorcycle), which [doesn’t have that bug](https://www.webpackbin.com/bins/-KsrH7tkqwwerDFy3ZMW).

staltz:
[...]

PR #668, opened 2017-08-24
fix(dom): wait for snabbdom's post hooks to clean vnodes (#667)

atomrc:
Initially, vnodes were cleaned (EventListener destroyed) during the snabbdom's \`remove\` or \`destroy\`
hook. Thus, if the vnode is, in fact, used latter on in the DOM, those listeners are now gone.

This patch only removes those nodes at the end of snabbdom's process. Thus if the nodes are needed
latter on, they won't be destroyed

ISSUES CLOSED: #667

### DIFF
--- a/dom/src/IsolateModule.ts
+++ b/dom/src/IsolateModule.ts
@@ -21,6 +21,13 @@ export class IsolateModule {
    */
   private fullScopesBeingUpdated: Array<string>;
 
+  /**
+   * A registry that keeps track of all the nodes that are removed from
+   * the virtual DOM in a single patch. Those nodes are cleaned once snabbdom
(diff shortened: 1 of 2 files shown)


atomrc:
side note: created a `pack` and tested it against my actual project, I can confirm it fixes #667 :)

Widdershin:
@atomrc sorry for the lack of update. I'll try to find some time to look at this on the weekend. Been moving house recently so hard to find time to focus.

staltz:
Good PR! Thanks a lot @atomrc !

atomrc:
My pleasure. Thanks for the review!

Any idea when could be the next release of @cycle/dom?

PR #671, opened 2017-08-29
feat(time): support creating custom operators with Time.createOperator()

Widdershin:
Also, rename scheduler.completion to scheduler.complete for consistency
with xstream addListener names.

Closes https://github.com/cyclejs/time/issues/31

### DIFF
--- a/time/most.ts
+++ b/time/most.ts
@@ -5,16 +5,11 @@ import {setAdapt} from '@cycle/run/lib/adapt';
 import {mockTimeSource as mockTimeSourceUntyped} from './src/mock-time-source';
 import {timeDriver as timeDriverUntyped} from './src/time-driver';
 import {Frame} from './src/animation-frames';
-import {Scheduler} from './src/scheduler';
+import {Comparator, OperatorArgs} from './src/types';
 
 setAdapt(stream => most.from(stream as any));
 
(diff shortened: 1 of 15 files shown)


time/most.ts
```diff
@@ -5,10 +5,16 @@ import {setAdapt} from '@cycle/run/lib/adapt';
 import {mockTimeSource as mockTimeSourceUntyped} from './src/mock-time-source';
 import {timeDriver as timeDriverUntyped} from './src/time-driver';
 import {Frame} from './src/animation-frames';
+import {Scheduler} from './src/scheduler';
 
 setAdapt(stream => most.from(stream as any));
 
 type Operator = <T>(stream: Stream<T>) => Stream<T>;
+export type Comparator = (actual: any, expected: any) => void;
```
jvanbruegge: Shouldn't it return `void | bool`?

Widdershin: 100% yes, good catch.

jvanbruegge:
I would also change the current operators to use the new `createOperator` method. This way it is easy to see how to do more complex operators by reading the source, plus it's a consistency win

Widdershin:
Good idea @jvanbruegge, I've done just that.

Widdershin:
@jvanbruegge thanks for the review and feedback, I really appreciate it! 😄

Issue #672, opened 2017-08-31
@cycle/time 0.9.1 Cannot find module './src/mock-time-source' from 'rxjs.js'


evenchange4:
**Code to reproduce the issue:**

There is a problem of module resolution when I update from `^0.8.0` to `^0.9.1`:

```js 
// error
import { mockTimeSource } from '@cycle/time/rxjs';
```

```
 ● Test suite failed to run
   Cannot find module './src/mock-time-source' from 'rxjs.js'
     at Resolver.resolveModule (../../node_modules/jest-resolve/build/index.js:179:17)
     at Object.<anonymous> (../../node_modules/@cycle/time/rxjs.js:7:26)
```

**Workaround**

Should I use the `rxjs.js ` file with specific path?

```js
// work!
import { mockTimeSource } from '@cycle/time/lib/rxjs';
```

Widdershin:
Hi @evenchange4, does this work correctly with `^0.9.2`?

Widdershin:
Hi @evenchange4 

Thanks for the issue!

This is an unintended consequence of the way the monorepo compiles packages. All code is outputted directly to the lib directory.

Previously I ran an additional compile step to ensure that `rxjs.js` and `most.js` were present in the root of the compiled package.

I would recommend your workaround in the meantime, and I'll chat to @staltz and @jvanbruegge to see if there's a way we can set this up so you can continue to import as before.

IssuehuntBot:
@issuehuntfest has funded $40.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/26316966/issues/672)

shriram-svg:
Opened https://github.com/cyclejs/cyclejs/pull/1003 to restore the generated root `rxjs`/`most` entrypoints and include them in the published package.

PR #673, opened 2017-08-31
chore(META): split @cycle/time into multiple packages

jvanbruegge:
ISSUES CLOSED: #672


jvanbruegge:
@Widdershin please review and dont forget to change the `dependencies` of all packages that require the common package before release :smirk:

jvanbruegge:
The test fails because the linking of the local package does not seem to work in the CI, running make test locally works. Should be resolved if the package is publish (maybe try with beta release first)

Widdershin:
@jvanbruegge 

Thanks for putting time and work into splitting this up.

I'm not sure yet that I want to split @cycle/time into separate packages like this.

I have already made mistakes in the past where I have forgotten to update the most and rxjs definitions, so I wanted to investigate generating them.

To that end, I have created #676 which generates es5 and es6 entry points for each stream library, and allows for easily supporting more compile targets and libraries in future.

It's a smaller change than this one, and I think it would be a bit less work to maintain.

Can you have a look at it and tell me what you think?

jvanbruegge:
@Widdershin That actually worked great! The auxilliary packages are completely generated now. No need to maintain them.

Widdershin:
Nice! What actually calls the generate packages script? Or do we need to do that manually when the type definitions change?

jvanbruegge:
postinstall, i could additionally call postinstall on lib.

Widdershin:
I think that's a good idea, or the postlib hook as in the other PR.

Do you ever sleep @jvanbruegge? Seems like you always get back to me straight away!

jvanbruegge:
Yes, I do sleep :D
But the night is simply at a different time here in germany.

Widdershin:
Mind = blown

jvanbruegge:
Done

staltz:
I hope I haven't asked this already, but why is there `time/common` and `time/xstream`? I would guess it should be `time` instead. (So in total there would be `time`, `time/rxjs`, `time/most`)

staltz:
I want to understand the underlying reason why this is deviating from the usual `run`, `rxjs-run`, `most-run` style.

jvanbruegge:
Well the `common` package is untyped. The postlib script then generates typed entry points for each stream lib.
If I would export the untyped versions from `common` I could adjust the generation script and rename `common`.

Widdershin:
I think my preference would be `@cycle/time` (xstream), `@cycle/rxjs-time` and `@cycle/most-time`. This makes it clear that `xstream` is the preferred option, and is consistent with the naming structure for `run`. In an ideal world I would have `rxjs` as a suffix but I think the inconsistency would irk me too much.

In any case I would love to move forward :)

Widdershin:
@jvanbruegge I can finish this off if you won't have a computer for a while, are you happy with that approach?

jvanbruegge:
Go for it :+1:

jvanbruegge:
Yay, new Laptop, so I can work again

staltz:
Okay, I'm finally reviewing this PR in details, and the `.scripts/generate-packages.js` approach is unexpected to me. I remember giving thumbs up for the postlib idea in #676:

```
"compile-entry-points-es5": "../node_modules/.bin/tsc --lib DOM,es6 --declaration xstream.ts most.ts rxjs.ts ",
"compile-entry-points-es6": "../node_modules/.bin/tsc --lib DOM,es6 --declaration --module es2015 --moduleResolution node es6.ts most-es6.ts rxjs-es6.ts",
"postlib": "npm run make-entry-points && npm run compile-entry-points-es5 && npm run compile-entry-points-es6"
```

But that doesn't involve source code generation, it's just a different compile target with `tsc`. On the other hand, `generate-packages.js` is clever but deviates quite a lot from what we do, e.g. with `@cycle/run`, `@cycle/most-run`, `@cycle/rxjs-run`. When I suggested `@cycle/time`, `@cycle/rxjs-time`, `@cycle/most-time` I was under the impression the approach would be the same as `-run` packages: one directory for `/time`, another for `/rxjs-time`, another for `most-time`. Right now we have `/time/rxjs` even though the npm package is eventually `@cycle/rxjs-time`.

If we had dozens of different stream libraries, I'd defend this `generate-packages.js` approach, but we have just 2 *other libraries* besides the main one, and to follow the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), the amount we have right now is not that too bad duplication to justify a generation step. 

> I have already made mistakes in the past where I have forgotten to update the most and rxjs definitions, so I wanted to investigate generating them.

I know this is a detail to remember when maintaining, but it is also important that the monorepo is coherent, with the same approach used everywhere. It helps bring predictability when maintaining or contributing to the project.

jvanbruegge:
> and to follow the Rule of Three

Technically we would follow the Rule with generation:
> It states that the code can be copied once, but that when the same code is used three times, it should be extracted into a new procedure

I really don't mind, I think we need @Widdershin input here

staltz:
As long as its coherent: source code generation used for most/rxjs everywhere applicable, or no source code generation anywhere. For now it would be best to stay coherent with the existing codebase. Looking into the future, too, I believe it's possible to remove most and rxjs related packages from Cycle.js if we leverage ES6 Proxy to do interception and conversion from xstream. It's too early to discuss that, but the point here is that even though we currently have code duplication for stream libraries, the tendency is to either keep only these 3, or to reduce to just xstream. We won't be adding more stream libraries.

indiescripter:
Hello cyclejs aficionados.

FWIW, I forked cyclejs a few weeks ago with a view to seeing how well [buildverse](https://www.npmjs.com/package/buildverse) and/or associated child project [for-each-project](https://www.npmjs.com/package/for-each-project) might be of use for cyclejs build infrastructure.

I managed to replace the make Makefile with npm scripts in an hour or so (using for-each-project utility in scripts) and felt that it was indeed a good, simple & effective fit.

Disclosure: Shameless self-promotion as I am the author of buildverse & for-each-project npm packages.

jvanbruegge:
@indiescripter Not really relevant to this discussion, please open an issue for that

jvanbruegge:
@staltz Done, removed the code gen

staltz:
That was quick. :)
I'm checking it.

jvanbruegge:
Forgot to reset the Makefile to the original state

most-time/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "lib/",
+    "dist/",
+    "most.js",
+    "rxjs.js",
+    "most.d.ts",
+    "rxjs.d.ts"
```
staltz: I think we don't need these `most` and `rxjs` files here

most-time/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@cycle/most-time",
+  "version": "0.10.0",
+  "description": "A time driver designed to enable awesome testing and dev tooling",
+  "license": "MIT",
+  "homepage": "https://cycle.js.org",
+  "bugs": "https://github.com/cyclejs/cyclejs/issues",
+  "repository": "https://github.com/cyclejs/cyclejs/tree/master/time",
```
staltz: `/time` => `/most-time`

most-time/.npmignore
```diff
@@ -0,0 +1,7 @@
+ignore/
+.idea/
+test/browser/page/tests-bundle.js
+*.sublime-project
+*.sublime-workspace
+most/
+rxjs/
```
staltz: Don't need `most/` and `rxjs/` here

most-time/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2016 Nick Johnstone
```
staltz: Since we're at it, might as well change it to 2016-2017

most-time/README.md
```diff
@@ -0,0 +1,3 @@
+## Only auxiliary package
+
+Please refer to [the main README](https://github.com/cyclejs/cyclejs/tree/master/time/common/README.md)
```
staltz: isn't the URL `...../cyclejs/cyclejs/tree/master/time/README.md`?

most-time/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "most.js",
+    "rxjs.js",
+    "most.d.ts",
+    "rxjs.d.ts"
+  ],
+  "browserify-shim": {
+    "most": "global:most"
+  },
+  "scripts": {
+    "start": "../../node_modules/.bin/budo -d example example/index.ts:index.js --live -- -p tsify",
```
staltz: We could remove `"start"` and this example. Anyway examples should live under the `/examples` directory.

most-time/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "test-watch": "../../node_modules/.bin/mocha 'test/**/*.ts' --compilers ts:ts-node/register --watch -R min",
+    "test-node": "../../node_modules/.bin/mocha 'test/**/*.ts' --compilers ts:ts-node/register",
+    "test": "npm run test-node && npm run test-docs",
+    "test-ci": "npm run test",
+    "test-docs": "markdown-doctest",
+    "browserify": "../../node_modules/.bin/browserify lib/index.js --global-transform=browserify-shim --standalone CycleTime --exclude xstream --outfile dist/cycle-time.js",
```
staltz: `--exclude xstream` => `--exclude most`

indiescripter:
Oh, sorry @jvanbruegge if my comment didn't seem relevant.

I've been following cyclejs GitHub emails for a few weeks now & thought this particular thread was rather apt given that
the title line is about multiple packages
and early on in this thread @staltz said re monorepo:
> I know this is a detail to remember when maintaining, but it is also important that the monorepo is coherent, with the same approach used everywhere. It helps bring predictability when maintaining or contributing to the project.

Nevertheless, I understand that this thread has evolved beyond initial discussion. I'd be happy to prepare a PR with what I envisage re cyclejs build process and/or as a separate issue if that would be a welcome contribution (at least for discussion).

rxjs-time/.npmignore
```diff
@@ -0,0 +1,7 @@
+ignore/
+.idea/
+test/browser/page/tests-bundle.js
+*.sublime-project
+*.sublime-workspace
+most/
+rxjs/
```
staltz: Remove `most/` and `rxjs/`

rxjs-time/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2016 Nick Johnstone
```
staltz: 2016-2017

rxjs-time/README.md
```diff
@@ -0,0 +1,3 @@
+## Only auxiliary package
+
+Please refer to [the main README](https://github.com/cyclejs/cyclejs/tree/master/time/common/README.md)
```
staltz: `....../tree/master/time/README.md`

rxjs-time/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@cycle/rxjs-time",
+  "version": "0.10.0",
+  "description": "A time driver designed to enable awesome testing and dev tooling",
+  "license": "MIT",
+  "homepage": "https://cycle.js.org",
+  "bugs": "https://github.com/cyclejs/cyclejs/issues",
+  "repository": "https://github.com/cyclejs/cyclejs/tree/master/time",
```
staltz: `..../tree/master/rxjs-time`

rxjs-time/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "lib/",
+    "dist/",
+    "most.js",
+    "rxjs.js",
+    "most.d.ts",
+    "rxjs.d.ts"
```
staltz: Remove `most` and `rxjs` files

rxjs-time/package.json
```diff
@@ -0,0 +1,50 @@
...
+  ],
+  "browserify-shim": {
+    "rxjs": "global:rxjs"
+  },
+  "scripts": {
+    "start": "../../node_modules/.bin/budo -d example example/index.ts:index.js --live -- -p tsify",
```
staltz: Remove `"start"`

rxjs-time/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "test-watch": "../../node_modules/.bin/mocha 'test/**/*.ts' --compilers ts:ts-node/register --watch -R min",
+    "test-node": "../../node_modules/.bin/mocha 'test/**/*.ts' --compilers ts:ts-node/register",
+    "test": "npm run test-node && npm run test-docs",
+    "test-ci": "npm run test",
+    "test-docs": "markdown-doctest",
+    "browserify": "../../node_modules/.bin/browserify lib/index.js --global-transform=browserify-shim --standalone CycleTime --exclude xstream --outfile dist/cycle-time.js",
```
staltz: `--exclude xstream` => `--exclude rxjs`

time/.npmignore
```diff
@@ -3,3 +3,5 @@ ignore/
 test/browser/page/tests-bundle.js
 *.sublime-project
 *.sublime-workspace
+most/
+rxjs/
```
staltz: Remove these

most-time/package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "files": [
+    "lib/",
+    "dist/"
+  ],
+  "browserify-shim": {
+    "most": "global:most"
+  },
+  "scripts": {
+    "test-watch": "../../node_modules/.bin/mocha 'test/**/*.ts' --compilers ts:ts-node/register --watch -R min",
```
staltz: These scripts should call `../node_modules/.bin/mocha` not `../../node_modules/.bin/mocha` now that the directory structure is different

staltz:
I can do this, by the way, unless you're already doing it

jvanbruegge:
Dumb copy+paste errors

staltz:
Actually, what should we do about tests for `rxjs-time` and `most-time`? They are all now in `time`. Note that with `rxjs-run` and `most-run` we have tests in their respective directories, even though they look similar to `run` tests.

jvanbruegge:
I would not copy+paste the same test file two times, if it is already working with all stream libraries.
I removed the tests from the *-time packages

staltz:
We need at least some kind of tests that make sure the package *works*. There can sometimes be bugs in the way the `package.json` files are written and we need to rely on Travis builds as an indication that everything is okay.

jvanbruegge:
Added one test per package

staltz:
And one more thing: the branch needs rebasing with `master` due to conflicts.

staltz:
@jvanbruegge 

```
TSError: ⨯ Unable to compile TypeScript
test/index.ts (3,24): Could not find a declaration file for module 'sinon'. '/home/travis/build/cyclejs/cyclejs/node_modules/sinon/lib/sinon.js' implicitly has an 'any' type.
  Try `npm install @types/sinon` if it exists or add a new declaration (.d.ts) file containing `declare module 'sinon';` (7016)
    at getOutput (/home/travis/build/cyclejs/cyclejs/node_modules/ts-node/src/index.ts:307:15)
    at /home/travis/build/cyclejs/cyclejs/node_modules/ts-node/src/index.ts:336:16
    at Object.compile (/home/travis/build/cyclejs/cyclejs/node_modules/ts-node/src/index.ts:496:11)
    at Module.m._compile (/home/travis/build/cyclejs/cyclejs/node_modules/ts-node/src/index.ts:392:43)
    at Module._extensions..js (module.js:579:10)
    at Object.require.extensions.(anonymous function) [as .ts] (/home/travis/build/cyclejs/cyclejs/node_modules/ts-node/src/index.ts:395:12)
```

Both for most-time and rxjs-time

PR #674, opened 2017-09-01
fix(http): allow string type for send property

evenchange4:
> By default sending strings will set the `Content-Type` to `application/x-www-form-urlencoded`.

Example from the document of superagent:

```js
  request.post('/user')
    // object
    .send({ name: 'tj' })
    // or string
    .send('name=tj')
    ...
```

Ref: http://visionmedia.github.io/superagent/

### DIFF
--- a/http/src/index.ts
+++ b/http/src/index.ts
@@ -18,7 +18,7 @@
  * the HTTP Source when querying for the response. E.g.
  * `sources.http.select(category)`
  * - `query` *(Object)*: an object with the payload for `GET` or `POST`.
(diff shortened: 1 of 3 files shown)


Widdershin:
Hi @evenchange4, thanks for the contribution!

Would you be able to add a simple test case for this functionality, just so that we don't accidentally regress it!

Thanks 😄

evenchange4:
@Widdershin Sure, I have push with a test case.

staltz:
Thanks @evenchange4 and thanks @Widdershin for the proactive behavior 👍 :)

PR #676, opened 2017-09-05
fix(time): fix importing most and rxjs with import '@cycle/time/rxjs'.

Widdershin:
Also allow importing es6 sources more easily.

ISSUES CLOSED: #672

Is an alternative to #673


jvanbruegge:
I still think it is suboptimal that the user decides which import he uses (especially for library authors).
But we could combine the two PRs. Multiple packages, but with generated typings.

Widdershin:
Are you referring to having to specify ES6 for rxjs and most?

jvanbruegge:
For xsteam too

jvanbruegge:
A library author then has to choose to either: screw treeshaking by using the default (!) or screw users that have tools without es6 features (like unmodified node)

Widdershin:
Oh yeah, that's a valid use case. Although ideally we want libraries built
around cycle time to be stream agnostic as well.

Have you got examples of possible libraries, so I can think it though?

Widdershin:
Also, if we have separate packages, which would a library author depend on? Just common? Struggling to see how this all fits together with multiple packages.

jvanbruegge:
The library author uses internally one version (xstream/rxjs/most) and uses adapt to marshal (just like normal cyclejs library). The thing he does not choose is cjs/es6. That will be done by the user consuming the library, i.e. the build tool of the user.

Widdershin:
That makes sense. Thanks for patiently explaining 😄

jvanbruegge:
I will later try to merge the two PRs, using your postlib code to generate the other packages

Widdershin:
Cool, I appreciate it! 👍

staltz:
@jvanbruegge I agree with the postlib idea. And @Widdershin, about "stream-agnostic", we used to try that, but we decided to just use xstream under the hood always.

jvanbruegge:
@Widdershin should we close this in favor of #673?

jvanbruegge:
Closing this as this will become irrelevant with the rewrite

PR #679, opened 2017-09-12
feat(time): Add Time.now() function to access current time as a Date

Widdershin:


danielchappell:
This is really great, thank you for responding to my question and adding this!

Widdershin:
👍 thanks for the reminder @danielchappell, had been meaning to do this for a while!

time/README.md
```diff
@@ -275,6 +275,15 @@ import {timeDriver, mockTimeSource} from '@cycle/time';
 ### `timeDriver()`
 The time driver returns a `TimeSource` object with the following methods:
 
+#### `now()`
+Returns the current time as a date. Useful for sampling time:
+
+```js
+const click$ = DOM.select('.button').events('click');
+
+const lastClickTime$ = click$.map(Time.now).remember();
```
jvanbruegge: maybe explicitly call `Time.now`, looks like a constant that way
```js
const lastClickTime$ = click$.map(() => Time.now()).remember(); //sample the current time on click
```

time/most.ts
```diff
@@ -18,6 +18,7 @@ interface TimeSource {
   throttle(period: number): Operator;
   periodic(period: number): Stream<number>;
   throttleAnimation: Operator;
+  now(): Date;
```
jvanbruegge: would wait for #673 and then generate those

time/README.md
witoldsz: If this lib is targeting the FP audience, the `() => x.fn()` looks terrible and could lead to confusion like: is it because it's shorter than `Time.now.bind(time)`, is it a method on some `Time` object??

Of course not, `Time.now` is just a function, what else would `stream$.map` accept?

Or maybe I am just picky after reading [_Chapter 2: First Class Functions_](https://mostly-adequate.gitbooks.io/mostly-adequate-guide/content/ch02.html) where author calls this an _ignorant_ vs _enlightened_:

```javascript
// ignorant
const getServerStuff = callback => ajaxCall(json => callback(json));

// enlightened
const getServerStuff = ajaxCall;
```

jvanbruegge:
Will be added in the rewrite

PR #682, opened 2017-09-23
feat(dom): overload .event() method using HTMLElementEventMap

ngfk:
The lib.dom.d.ts file contains a mapping where the name of the event type is mapped to the event
type itself (e.g. click -> MouseEvent). Using this mapping on the DOMSource.events() function allows
the correct event stream to be returned.

### DIFF
--- a/dom/src/DOMSource.ts
+++ b/dom/src/DOMSource.ts
@@ -8,5 +8,9 @@ export interface EventsFnOptions {
 export interface DOMSource {
   select(selector: string): DOMSource;
   elements(): MemoryStream<Document | Element | Array<Element> | string>;
+  events<K extends keyof HTMLElementEventMap>(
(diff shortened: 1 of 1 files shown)


jvanbruegge:
That's cool, I didn't knew that HTMLElementEventMap existed

jvanbruegge:
And overloading is the right move, because we might want to use custom events

jvanbruegge:
Thanks for the PR!!

ngfk:
No problem, happy to help!

staltz:
Wow, HTMLElementEventMap is great! Thanks for the PR

donaldpipowitch:
This is so cool. Just searched for this. Thank you for the PR. Looking forward to a release.

SteveALee:
Sweet. Thank you

PR #683, opened 2017-09-25
refactor(examples): updated nested folders example

usm4n:
Updated the nested folders example to use the new Collection API from the Onionify 4.0 Library.

### DIFF
--- a/examples/nested-folders/src/Folder.js
+++ b/examples/nested-folders/src/Folder.js
@@ -1,7 +1,7 @@
 import xs from 'xstream'
 import isolate from '@cycle/isolate'
 import {div, button} from '@cycle/dom'
-import {pick, mix} from 'cycle-onionify'
+import {makeCollection} from 'cycle-onionify'
 
 function generateId() {
   return Number(String(Math.random()).replace(/0\.0*/, ''))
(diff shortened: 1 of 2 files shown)


examples/nested-folders/src/Folder.js
```diff
@@ -78,29 +78,19 @@ function view(state$, childrenVDOM$) {
...
-    .compose(mix(xs.combine))
-
-  return {
-    DOM: childrenVDOM$,
-    onion: childrenReducer$,
-  }
-}
-
 export default function Folder(sources) {
+  const Children = makeCollection({
```
jvanbruegge: I would leave `Children` outside of Folder

staltz:
Hi @usm4n :)
Thanks for submitting your first PR, we appreciate any kind of initiative to help. Before merging this PR could you notice the comment @jvanbruegge placed? I think I agree and it's the only thing to change before merging.

examples/nested-folders/src/Folder.js
usm4n: @jvanbruegge I wrote it like this:
~~~
function Children(sources = {}) {
  const children = makeCollection({
    item: Folder,
    itemKey: state => state.id,
    itemScope: key => key,
    collectSinks: instances => ({
      onion: instances.pickMerge('onion'),
      DOM: instances.pickCombine('DOM')
    })
  })

  return children
}

export default function Folder(sources) {
  const childrenSinks = isolate(Children(), 'children')(sources)
  ...
~~~
Then had kind of mixed feelings since I just started learning this awesome framework. So I thought it would make sense to move the code inside the `Folder` to highlight the implicit creation of the `Children` component.

examples/nested-folders/src/Folder.js
```diff
@@ -78,29 +78,23 @@ function view(state$, childrenVDOM$) {
...
-  )
-
-  const childrenReducer$ = childrenSinks$
-    .compose(pick('onion'))
-    .compose(mix(xs.merge))
-
-  const childrenVDOM$ = childrenSinks$
-    .compose(pick('DOM'))
-    .compose(mix(xs.combine))
+function Children(sources = {}) {
```
jvanbruegge: The function wrapper is not needed here: You have:
```
function X(something) {
    const somethingElse = ...;
    return somethingElse(something)
}
```
I would remove that wrapper:
```js
const Children = makeCollection({
  /* ... */
});
```

usm4n:
@staltz thank your for the thumbs up.

@jvanbruegge I have made the proposed changes, thanks for the guidance :)

jvanbruegge:
All good to go, just run `make prettier-all` once please.

staltz:
@jvanbruegge I think prettier is run automatically with the git commit. Also, before merging, we could squash all the commits into one.

jvanbruegge:
Maybe if the commit was done with `make commit` but the second one wasnt. There is a semicolon missing.

usm4n:
Ok, I have squashed the commits after running the `prettier` command. I hope everything is fine now.

jvanbruegge:
Almost ^^
Can you remove the http and time file from the commit (I assume you used `git add -a`)
Just do
```
git reset HEAD~1
git add examples/
make commit
```

jvanbruegge:
Thanks for the PR!

Issue #681, opened 2017-09-19
Implement `ok` field in request objects


staltz:
Cycle/HTTP should support an `ok` field to determine that some responses like 4xx or 5xx should be treated as normal responses, not errors:

```js
{
  url: 'http://.......',
  method: 'GET',
  category: 'mycat',
  ok: res => res.status < 500,
}
```

See comment https://github.com/cyclejs/cyclejs/issues/669#issuecomment-329949712

jvanbruegge:
Will see if I can do that later

PR #684, opened 2017-09-27
feat(http): implement `ok` field in request objects

nlarche:
ISSUES CLOSED: #681

- [X] I added new tests for the issue I fixed/built
- [X] I ran `make test FOO` for the package FOO I'm modifying

Hello, 

I only add one test for this feature, tell me if it's enough.

I hope this help.

Nicolas

### DIFF
--- a/http/src/interfaces.ts
+++ b/http/src/interfaces.ts
@@ -31,6 +31,7 @@ export interface RequestOptions {
   category?: string;
   lazy?: boolean;
   responseType?: string;
+  ok?: ((res: SuperagentResponse) => boolean);
   _error?: any;
   _namespace?: Array<string>;
 }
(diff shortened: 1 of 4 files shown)


http/src/interfaces.ts
```diff
@@ -31,6 +31,7 @@ export interface RequestOptions {
   category?: string;
   lazy?: boolean;
   responseType?: string;
+  ok?: Function;
```
jvanbruegge: I would change that to `(res: SuperAgentResponse) => boolean` to make the typings better

nlarche: Ok I change that. I'm not very familiar with typescript yet.

staltz:
Travis tests pass too. In my opinion this is ready to merge. :)

jvanbruegge:
:tada:

PR #688, opened 2017-09-28
Simplify error handling API of cycle/http

bdc34:
The API of cycle/http conflates HTTP responses with error status and
xstream errors. Makes users of cycle/http use two different API
mechanisms from xstream to work with cycle/http. Normal xstream events
for HTTP responses with status codes in the 200 range and xstream errors
for HTTP responses outside of the 200 range. Timeouts etc use xstream
errors too.

The solution I pursued was to make all responses from Superagent sent as
normal xstream events. So now everything that comes out of cycle/http is
a normal xstream event and can be filtered in basic xstream
ways. replaceError() never needs to be used and no considerations for if
streams have been replaced needs to be made. Diamond paterns can still
be used if desired. If the user of the library wants xstream errors they
can do that too.

This makes the API of cycle/http better adhere to the principle of least
supprise. It also simplifis it since there is only one xstream API
mechanism that needs to be used instead of two.

### DIFF
--- a/http/src/http-driver.ts
+++ b/http/src/http-driver.ts
@@ -100,16 +100,24 @@ export function createResponse$(reqInput: RequestInput): Stream<Response> {
           });
         }
         this.request.end((err: any, res: Response) => {
-          if (err) {
+          var out;
+          if (err && ! res) {
             if (err.response) {
               err.response.request = reqOptions;
+              out = err.response;              
+              out.error = err;              
+              delete out.err.response;//throw out cicrular reference
+            }else{
+             //COMMENT ONLY FOR PULL REQUEST: No need to add 'error?: any' to interface.ts
+             // because it is already a property of SuperagentResponse
+              out = {error: err, request:reqOptions};
             }
-            listener.error(err);
           } else {
             res.request = reqOptions;
-            listener.next(res);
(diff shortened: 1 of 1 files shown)


geovanisouza92:
This has been resolved in #681 and would be a breaking change

bdc34:
Looks great. What does #681 this do in the cases of timeout of hostname lookup failure?

geovanisouza92:
I'm not sure, but probably emit an `listener.error`, as the previous behavior. What was changed is the ability to classify "soft errors" (when you have a HTTP response with status code != 200, e.g.).

staltz:
Hi @bdc34  :)
Thanks for submitting this PR, we appreciate any kind of desire to help. Like @geovanisouza92 said, Jan's PR for #681 should have solved this by now. We still need to issue a new version for that, though. I can do that, now/soon.

bdc34:
@staltz thanks for the encouragement. Please consider simplifying the cycle/http API with a breaking change some time in the future if you do plan to make major changes. I think having all http activity come out as normal xstream events from cycle/http will be less surprising to newcomers.

staltz:
New cycle/http version: v14.4.0 released

PR #689, opened 2017-09-29
refactor(dom): use native ES6 Map

jvanbruegge:
As stated in the README of es6-map, the algorithmic complexity of the poly-/ponyfill is O(n) not
O(1). As almost all browsers have native ES6 Map support nowadays, we should use them and only
polyfill if needed


staltz:
Perfect PR, thanks @jvanbruegge

PR #690, opened 2017-09-30
Remove "change" from eventTypesThatDontBubble

abaco:
This fixes an issue I had where a parent component would receive a `change` event from an isolated child component. It seems a trivial bug: `change` is treated as non-bubbling while in fact [it does bubble](https://developer.mozilla.org/en-US/docs/Web/Events/change#General_info).

### DIFF
--- a/dom/src/MainDOMSource.ts
+++ b/dom/src/MainDOMSource.ts
@@ -18,7 +18,6 @@ const eventTypesThatDontBubble = [
   `blur`,
   `canplay`,
   `canplaythrough`,
-  `change`,
   `durationchange`,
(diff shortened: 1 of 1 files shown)


jvanbruegge:
Thanks @abaco, will merge after the tests pass

staltz:
Reminder: this is actually a breaking change, because if someone was unknowingly expecting or depending on the fact that the parent gets a `change` event from an isolated child, their app may break in production. Breaking change is anything that may cause a noticeably different behavior of the application that uses the library.

I can update the commit message before publishing a new version.

Issue #677, opened 2017-09-11
Make DOMSource.elements() always return array, create .element() for single element


Widdershin:
Currently, in the DOM driver, `DOM.select('.foo').elements()` has a type of `Element | Element[]`.

This is because when selecting the root element, the document or the body, there is only one possible result.

However, this has poor user experience, as it means users have to coerce the type depending on what they are selecting, which means we might as well not have types for this usecase.

Additionally, it is common to want to take the first item from the resulting array, in which case you must do `DOM.select('.foo').elements().filter(elements => elements > 0).map(elements => elements[0])`.

I propose to instead have `.elements()` always return an `Element[]`, and to create an alternative `.element()` function that always returns an `Element`.

This means that users would not have to coerce the type, and could instead use the function appropriate for their use case.

`.element()` would be effectively equivalent to `.elements().filter(elements => elements > 0).map(elements => elements[0])`, which would save having to include that relatively nasty looking code.

I don't think this would make the user experience worse in any way, but would make our type information more useful and our code elegant and more readable.

The biggest downside is that this is a breaking change, as I am proposing that when selecting root, document or body, and using `.elements()`, we return a single item array.

staltz:
Yes, I think we should just do this. In the case of `.elements()` for the document or body, it's okay to return an array with one entry.

PR #691, opened 2017-10-01
refactor(dom): make DOMSource.elements() monomorphic

jvanbruegge:
MainDOMSource.elements() now always returns an array of Elements, return types of the other
DOMSources are correctly inferred

ISSUES CLOSED: #677

### DIFF
--- a/dom/src/DOMSource.ts
+++ b/dom/src/DOMSource.ts
@@ -7,7 +7,9 @@ export interface EventsFnOptions {
 
 export interface DOMSource {
   select(selector: string): DOMSource;
-  elements(): MemoryStream<Document | Element | Array<Element> | string>;
+  elements(): MemoryStream<
+    Document | HTMLBodyElement | Array<Element> | string
+  >;
   events<K extends keyof HTMLElementEventMap>(
     eventType: K,
     options?: EventsFnOptions,
--- a/dom/src/MainDOMSource.ts
+++ b/dom/src/MainDOMSource.ts
@@ -88,6 +88,11 @@ function filterBasedOnIsolation(domSource: MainDOMSource, fullScope: string) {
   };
 }
 
+export interface SpecialSelector {
+  body: BodyDOMSource;
+  document: DocumentDOMSource;
+}
+
 export class MainDOMSource implements DOMSource {
   constructor(
     private _rootElement$: Stream<Element>,
@@ -111,10 +116,10 @@ export class MainDOMSource implements DOMSource {
     };
   }
 
-  public elements(): MemoryStream<Element> {
-    let output$: Stream<Element | Array<Element>>;
+  public elements(): MemoryStream<Array<Element>> {
+    let output$: Stream<Array<Element>>;
     if (this._namespace.length === 0) {
-      output$ = this._rootElement$;
+      output$ = this._rootElement$.map(x => [x]);
     } else {
       const elementFinder = new ElementFinder(
         this._namespace,
@@ -122,7 +127,7 @@ export class MainDOMSource implements DOMSource {
       );
       output$ = this._rootElement$.map(el => elementFinder.call(el));
     }
-    const out: DevToolEnabledSource & MemoryStream<Element> = adapt(
+    const out: DevToolEnabledSource & MemoryStream<Array<Element>> = adapt(
       output$.remember(),
     );
     out._isCycleSource = this._name;
@@ -133,6 +138,10 @@ export class MainDOMSource implements DOMSource {
     return this._namespace;
   }
 
+  public select<T extends keyof SpecialSelector>(
+    selector: T,
+  ): SpecialSelector[T];
+  public select(selector: string): MainDOMSource;
   public select(selector: string): DOMSource {
     if (typeof selector !== 'string') {
       throw new Error(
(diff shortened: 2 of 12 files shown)


jvanbruegge:
The coolest thing about this PR is here:
https://github.com/cyclejs/cyclejs/pull/691/files#diff-c6eab028a43003b501fa867456f9388cR50
https://github.com/cyclejs/cyclejs/pull/691/files#diff-c6eab028a43003b501fa867456f9388cR78
Those tests compile without a problem! If you change the type annotation they wont.

jvanbruegge:
No idea why the tests fail, they pass on my machine

staltz:
@jvanbruegge this is a promising PR, but we need to polish some corners before merging.

> No idea why the tests fail, they pass on my machine

If you look at the Travis CI output, you should see some `not ok 105 IE 10.0` results when testing cycle/dom. Tests should pass on your machine because you're probably not testing IE 10. I'll comment in-place in the source code soon.

dom/src/DOMSource.ts
```diff
@@ -1,13 +1,23 @@
...
   useCapture?: boolean;
   preventDefault?: PreventDefaultOpt;
 }
 
-export interface DOMSource {
+export type DOMSource =
+  | MainDOMSource
+  | DocumentDOMSource
+  | BodyDOMSource
+  | MockedDOMSource;
```
staltz: Is the union type here necessary to achieve this PR? I was intending to eventually place JSDocs in the methods of the `DOMSource` interface, so that docs for `elements`, `events`, `select` etc would show when hovering in VSCode or elsewhere in other IDEs. Currently, docs for these methods are dumped into the JSDoc for `makeDOMDriver` in `index.ts`.

With a union type, it's unfortunate we can't provide the JSDoc for each method, and people will be importing `import {DOMSource} from '@cycle/dom'` instead of `import {MainDOMSource} from '@cycle/dom'`

dom/src/DOMSource.ts
```diff
@@ -1,13 +1,23 @@
...
+
+export interface BaseDOMSource {
   select(selector: string): DOMSource;
-  elements(): MemoryStream<Document | Element | Array<Element> | string>;
+  elements(): MemoryStream<Document | HTMLBodyElement | Array<Element>>;
```
staltz: I'm afraid the removal of the `string` case would break `cycle/html` which has `HTMLSource implements DOMSource` and `elements(): MemoryStream<string>`

dom/src/MainDOMSource.ts
```diff
@@ -111,18 +116,18 @@ export class MainDOMSource implements DOMSource {
     };
   }
 
-  public elements(): MemoryStream<Element> {
-    let output$: Stream<Element | Array<Element>>;
+  public elements(): MemoryStream<Array<Element>> {
+    let output$: Stream<Array<Element>>;
     if (this._namespace.length === 0) {
-      output$ = this._rootElement$;
+      output$ = this._rootElement$.map(Array.of);
```
staltz: `Array.of` is ES6, and I suspect this is what is breaking tests on IE10. We could just use `.map(x => [x])` instead.

dom/src/DOMSource.ts
jvanbruegge: AFAIK it is needed for the type inference to work. But I have to check this

dom/src/DOMSource.ts
jvanbruegge: I see

dom/src/MainDOMSource.ts
jvanbruegge: ah, yes of course

jvanbruegge:
Will fix those issues tomorrow

dom/src/DOMSource.ts
jvanbruegge: Was not needed, removed the union

dom/src/DOMSource.ts
jvanbruegge: readded string, did not interfer with type inference

dom/src/DOMSource.ts
```diff
@@ -1,13 +1,19 @@
 import {Stream, MemoryStream} from 'xstream';
 import {PreventDefaultOpt} from './fromEvent';
+import {MainDOMSource} from './MainDOMSource';
+import {DocumentDOMSource} from './DocumentDOMSource';
+import {BodyDOMSource} from './BodyDOMSource';
+import {MockedDOMSource} from './mockDOMSource';
```
staltz: We don't need these imports anymore

staltz:
@jvanbruegge another very important thing is to reword one of the commits so it has a `BREAKING CHANGE:` section to be shown in the changelog since this issue is a breaking change.

staltz:
@jvanbruegge, too fast. I didn't mark this with PR: 2 LGTM.

In fact I was just about to write some comments: we need commit `fix(dom)` not `refactor(dom)` and it's still missing updates in the JSDocs. I also think the breaking changes note can be clearer: `BodyDOMSource.elements()` and `DocumentDOMSource.elements()` return stream of Node, but `MainDOMSource.elements()` returns Stream of Array of Node.

To revert this, recreate the branch, cherry-pick this commit, then submit a new PR. I'll rollback this merged commit after that.

jvanbruegge:
sorry

Issue #685, opened 2017-09-27
@cycle/http probably wrong isolation function?


baranovxyz:
```request._namespace.indexOf(scope) !== -1```
I mean what if I isolated some component twice with same scope name?
```
export const isolateSource = (
  source: DriverSource,
  scope: string | null
): DriverSource =>
  scope === null
    ? source
    : source.filter(
        (request: RequestOptions) =>
          Array.isArray(request._namespace) &&
          request._namespace.indexOf(scope) !== -1
      );
```

jvanbruegge:
I created [a test](https://github.com/cyclejs/cyclejs/pull/692/files#diff-83140a82268528effa0d84ac231ef446R456) that would indeed trigger unwanted behavior without the fix.

staltz:
Fixed by Jan, and new version cycle/http v14.4.0 released!

PR #692, opened 2017-10-01
fix(http): check position of isolation namespace

jvanbruegge:
ISSUES CLOSED: #685


staltz:
Good PR! Thanks @jvanbruegge

Issue #595, opened 2017-04-29
Document that `isolate` is not type-transparent


mightyiam:
The issue described in #559 should be documented. Because if the user is unaware, this turns his carefully crafted types into `any`s, silently. And it is considered explicit. So `noImplicitAny` does not help. I remember that before becoming aware of this, I had to go through some code flow to discover where my types are disappearing.

It should be documented in https://cycle.js.org/api/isolate.html.

mightyiam:
This is one of those onboarding pains.

PR #695, opened 2017-10-03
docs(isolate): document isolate as non type transparent

xtianjohns:
ISSUES CLOSED: #595


staltz:
Thanks @xtianjohns ! Easy to merge :)

Issue #694, opened 2017-10-02
Get source maps right in all packages


mightyiam:
What I've described [in xstream](https://github.com/staltz/xstream/issues/221) is probably going on in at least some of the packages here. At least for `@cycle/history`, I get this:

```
WARNING in ./node_modules/@cycle/history/lib/es6/captureClicks.js
(Emitted value instead of an instance of Error) Cannot find source file '../../src/captureClicks.ts': Error: Can't resolve '../../src/captureClicks.ts' in '/home/s
hahar/src/cyclejs-components/www/node_modules/@cycle/history/lib/es6'            
 @ ./node_modules/@cycle/history/lib/es6/index.js 75:0-48                                                                                                           @ ./src/index.ts                                                                
                                                                                 
WARNING in ./node_modules/@cycle/history/lib/es6/createHistory$.js
(Emitted value instead of an instance of Error) Cannot find source file '../../src/createHistory$.ts': Error: Can't resolve '../../src/createHistory$.ts' in '/home
/shahar/src/cyclejs-components/www/node_modules/@cycle/history/lib/es6'
 @ ./node_modules/@cycle/history/lib/es6/drivers.js 2:0-50                       
 @ ./node_modules/@cycle/history/lib/es6/index.js                                                                                                                   @ ./src/index.ts                                                                
                                                                                 
WARNING in ./node_modules/@cycle/history/lib/es6/drivers.js
(Emitted value instead of an instance of Error) Cannot find source file '../../src/drivers.ts': Error: Can't resolve '../../src/drivers.ts' in '/home/shahar/src/cy
clejs-components/www/node_modules/@cycle/history/lib/es6'                        
 @ ./node_modules/@cycle/history/lib/es6/index.js 37:0-46 54:0-50 93:0-52                                                                                           @ ./src/index.ts                       
                                        
WARNING in ./node_modules/@cycle/history/lib/es6/index.js
(Emitted value instead of an instance of Error) Cannot find source file '../../src/index.ts': Error: Can't resolve '../../src/index.ts' in '/home/shahar/src/cyclej
s-components/www/node_modules/@cycle/history/lib/es6'
 @ ./src/index.ts 4:18-43
```

mightyiam:
Seems I've already reported this in #577. Thanks for working on it in #696.

staltz:
I'll be releasing new minor versions for each package, soon

mightyiam:
Awesome stuff! Thank you!

PR #696, opened 2017-10-03
chore(META): make relative paths of cjs and es6 build the same

jvanbruegge:
ISSUES CLOSED: #694

### DIFF
--- a/run/src/adapt.ts
+++ b/run/src/adapt.ts
@@ -3,20 +3,18 @@ import {Stream} from 'xstream';
 declare var window: any;
 
 function getGlobal(): any {
-  let globalObject: any;
+  let globalObj: any;
   if (typeof window !== 'undefined') {
-    globalObject = window;
+    globalObj = window;
   } else if (typeof global !== 'undefined') {
-    globalObject = global;
+    globalObj = global;
   } else {
-    globalObject = this;
+    globalObj = this;
   }
-
-  globalObject.Cyclejs = globalObject.Cyclejs || {};
-  globalObject.Cyclejs.adaptStream =
-    globalObject.Cyclejs.adaptStream || ((x => x) as AdaptStream);
-
-  return globalObject;
+  globalObj.Cyclejs = globalObj.Cyclejs || {};
+  globalObj = globalObj.Cyclejs;
+  globalObj.adaptStream = globalObj.adaptStream || ((x => x) as AdaptStream);
+  return globalObj;
 }
 
 export interface AdaptStream {
@@ -24,9 +22,9 @@ export interface AdaptStream {
 }
 
 export function setAdapt(f: AdaptStream): void {
-  getGlobal().Cyclejs.adaptStream = f;
+  getGlobal().adaptStream = f;
 }
 
 export function adapt(stream: Stream<any>): any {
-  return getGlobal().Cyclejs.adaptStream(stream);
+  return getGlobal().adaptStream(stream);
 }
--- a/run/src/adapt.ts
+++ b/run/src/adapt.ts
@@ -1,15 +1,32 @@
 import {Stream} from 'xstream';
 
+declare var window: any;
+
+function getGlobal(): any {
+  let globalObject: any;
+  if (typeof window !== 'undefined') {
+    globalObject = window;
+  } else if (typeof global !== 'undefined') {
+    globalObject = global;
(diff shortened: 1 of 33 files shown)


jvanbruegge:
This PR contains basicly three things:
- Compile cjs to `lib/cjs` and not `lib` to keep the path for the source maps the same
- Add `adaptStream` to global object instead of free variable (needed to avoid a breaking change when importing `@cycle/run/lib/adapt`
- Fix import paths in the tests

run/src/adapt.ts
```diff
@@ -4,12 +4,12 @@ export interface AdaptStream {
   (s: Stream<any>): any;
 }
 
-let adaptStream: AdaptStream = x => x;
+(global as any).adaptStream = (x => x) as AdaptStream;
```
staltz: We should be careful with the `global` keyword: https://github.com/tc39/proposal-global see also http://kangax.github.io/compat-table/esnext/#test-global

Instead, let's do something like this https://github.com/ReactiveX/rxjs/blob/c7c337e134605c9629ccfcaaf7ef562cd2bb9c18/src/Rx.global.js

run/src/adapt.ts
```diff
@@ -1,15 +1,24 @@
 import {Stream} from 'xstream';
 
+declare var window: any;
+
+function getGlobal(): any {
+  if(typeof window !== undefined) {
+    return window;
+  }
+  return global;
+}
```
staltz: `typeof window !== undefined` will always be true because it should be `typeof window !== 'undefined'` instead. Also we should do the same typeof check for `global`, returning `this` as the last option.

run/src/adapt.ts
```diff
@@ -1,15 +1,27 @@
...
+  }
+  return this;
+}
+
 export interface AdaptStream {
   (s: Stream<any>): any;
 }
 
-let adaptStream: AdaptStream = x => x;
+getGlobal().adaptStream = (x => x) as AdaptStream;
```
staltz: Actually I hadn't realized this before, but since the singleton `adaptStream` is now attached to a shared global, e.g. `window`, we need to be careful with naming (unlike before). I'd recommend something likely to be unique, like `window.Cyclejs.adaptStream` since cycle/run already attaches `window.Cyclejs.sinks` for the purposes of the DevTool.

run/src/adapt.ts
```diff
@@ -1,15 +1,27 @@
...
+  }
+  return this;
+}
+
 export interface AdaptStream {
   (s: Stream<any>): any;
 }
 
-let adaptStream: AdaptStream = x => x;
+getGlobal().Cyclejs.adaptStream = (x => x) as AdaptStream;
```
staltz: `getGlobal().Cyclejs` could be undefined in which case we need to check and create it. I'd recommend actually putting this `Cyclejs` container inside the `getGlobal()` function, doing the conditional create there. 

Check e.g. 
https://github.com/cyclejs/cyclejs/blob/8deedf9c91f187f77b2506a0e859e472526b7056/run/src/index.ts#L223-L224

jvanbruegge:
@staltz Anything left to do?

staltz:
@jvanbruegge I did a simple refactor of adapt.ts, just to avoid object property lookups like `globalObj.Cyclejs`, by making the globalObj be that `globalObj.Cyclejs` itself.

About the HTTP bug, I think it's just TypeScript complaining on some test files. I can put a commit for that as well, wouldn't bother that it's under this branch because it's refactor of tests.

jvanbruegge:
Ok

staltz:
In this case it was very useful to fix TypeScript complaints because it turned out to show me a bug with `http/rxjs-typings.d.ts`, see latest commit. I'll wait for Travis builds now

jvanbruegge:
@staltz can you merge? We would need new version for all packages

staltz:
This is marked `work in progress` because I want to first release a major version of cycle/dom with #709 before releasing cycle/dom major with #690. That is, let's bundle two major versions of cycle/dom together so that migration is easier. After that, we can merge this PR and release minor versions for all packages.

staltz:
Oh, one thing that's missing in this PR is updating `.scripts/make-api-docs.js`

staltz:
Waiting for Travis CI to become green.

PR #706, opened 2017-10-05
fix(time): ignore node.js specific modules in React Native

staltz:
I'm using cycle/time in a demo React Native app, and noticed that it won't run because `chalk` (used by `variable-diff`) expects a node.js environment. This PR ignores node.js-specific modules like that one.


staltz:
Thanks, I'll merge and make `0.10.1`

PR #707, opened 2017-10-05
docs (examples): fix link

blakelapierre:
Looks like entire directory was moved in commit 41d7759302d474779ce8888a3b9cc09924681966

Other links could be broken...

I think I messed up the commit message. You may want to change it or make a new commit.


staltz:
Hi @blakelapierre  :)
Thanks for the PR. If you don't mind I'll do this fix in another commit because it should be done on the file `/docs/content/index.md` instead. Thanks anyway!

PR #709, opened 2017-10-06
fix(dom): make DOMSource.elements() monomorphic

jvanbruegge:
MainDOMSource.elements() now always returns an array of Elements, return types of the other
DOMSources are correctly inferred

ISSUES CLOSED: #677
BREAKING CHANGE: `elements()` now always returns an array of elements if
you pass a selector

### DIFF
--- a/dom/src/DOMSource.ts
+++ b/dom/src/DOMSource.ts
@@ -7,7 +7,10 @@ export interface EventsFnOptions {
 
 export interface DOMSource {
   select(selector: string): DOMSource;
-  elements(): MemoryStream<Document | Element | Array<Element> | string>;
+  elements(): MemoryStream<
+    Array<Document> | Array<HTMLBodyElement> | Array<Element> | string
+  >;
+  element(): MemoryStream<Document | HTMLBodyElement | Element>;
   events<K extends keyof HTMLElementEventMap>(
     eventType: K,
     options?: EventsFnOptions,
--- a/dom/src/DocumentDOMSource.ts
+++ b/dom/src/DocumentDOMSource.ts
@@ -7,12 +7,20 @@ import {fromEvent} from './fromEvent';
 export class DocumentDOMSource implements DOMSource {
   constructor(private _name: string) {}
 
-  public select(selector: string): DOMSource {
+  public select(selector: string): DocumentDOMSource {
     // This functionality is still undefined/undecided.
     return this;
   }
 
-  public elements(): MemoryStream<Document> {
+  public elements(): MemoryStream<Array<Document>> {
+    const out: DevToolEnabledSource & MemoryStream<Array<Document>> = adapt(
+      xs.of([document]),
+    );
+    out._isCycleSource = this._name;
+    return out;
+  }
+
+  public element(): MemoryStream<Document> {
     const out: DevToolEnabledSource & MemoryStream<Document> = adapt(
       xs.of(document),
     );
--- a/dom/src/MainDOMSource.ts
+++ b/dom/src/MainDOMSource.ts
@@ -87,6 +87,11 @@ function filterBasedOnIsolation(domSource: MainDOMSource, fullScope: string) {
   };
 }
 
+export interface SpecialSelector {
+  body: BodyDOMSource;
+  document: DocumentDOMSource;
+}
+
 export class MainDOMSource implements DOMSource {
   constructor(
     private _rootElement$: Stream<Element>,
@@ -110,10 +115,10 @@ export class MainDOMSource implements DOMSource {
     };
   }
 
-  public elements(): MemoryStream<Element> {
-    let output$: Stream<Element | Array<Element>>;
+  public elements(): MemoryStream<Array<Element>> {
+    let output$: Stream<Array<Element>>;
     if (this._namespace.length === 0) {
-      output$ = this._rootElement$;
+      output$ = this._rootElement$.map(x => [x]);
     } else {
       const elementFinder = new ElementFinder(
         this._namespace,
@@ -121,9 +126,19 @@ export class MainDOMSource implements DOMSource {
       );
       output$ = this._rootElement$.map(el => elementFinder.call(el));
     }
-    const out: DevToolEnabledSource & MemoryStream<Element> = adapt(
+    const out: DevToolEnabledSource & MemoryStream<Array<Element>> = adapt(
       output$.remember(),
     );
+    out._isCycleSource = this._name;
+    return out;
+  }
+
+  public element(): MemoryStream<Element> {
+    const output$: MemoryStream<Element> = this.elements()
+      .filter(arr => arr.length > 0)
+      .map(arr => arr[0])
+      .remember();
+    const out: DevToolEnabledSource & MemoryStream<Element> = adapt(output$);
     out._isCycleSource = this._name;
     return out;
   }
@@ -132,6 +147,10 @@ export class MainDOMSource implements DOMSource {
     return this._namespace;
   }
 
+  public select<T extends keyof SpecialSelector>(
+    selector: T,
+  ): SpecialSelector[T];
+  public select(selector: string): MainDOMSource;
   public select(selector: string): DOMSource {
     if (typeof selector !== 'string') {
       throw new Error(
@@ -156,7 +175,7 @@ export class MainDOMSource implements DOMSource {
       this._isolateModule,
       this._delegators,
       this._name,
-    );
+    ) as DOMSource;
   }
 
   public events(
--- a/dom/src/index.ts
+++ b/dom/src/index.ts
@@ -11,10 +11,12 @@ export {MainDOMSource} from './MainDOMSource';
  * Snabbdom "VNode" objects. The output of this driver is a "DOMSource": a
  * collection of Observables queried with the methods `select()` and `events()`.
  *
- * `DOMSource.select(selector)` returns a new DOMSource with scope restricted to
- * the element(s) that matches the CSS `selector` given.
+ * **`DOMSource.select(selector)`** returns a new DOMSource with scope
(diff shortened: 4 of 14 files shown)


staltz:
These are to do:

- commit message `refactor(dom): ...` to `fix(dom): ...`
- update `src/index.ts` JSDocs
- clear breaking changes instruction in the commit message

That said, now that I took a second look at issue #677, I think the proposal is to introduce *also* a new API: `element()`, and then make `elements()` truly return `Stream<Array<*>>` for all cases, including BodyDOMSource and DocumentDOMSource.

jvanbruegge:
> I think the proposal is to introduce also a new API: 

I think this is not needed, because a call to `DOM.select('document')` gets automaticly infered to `DocumentDOMSource` and the return type is clear then.

fuunnx:
I agree with Staltz, I find this is more intuitive because `elements` is plural. Following the principle of least surprise I think it should always return an Array (especially important for non ts users).

staltz:
@jvanbruegge Thanks for the updates, I think it's mostly good now, but to still talk about the `element` proposal, Nick initially suggested:

> Additionally, it is common to want to take the first item from the resulting array, in which case you must do `DOM.select('.foo').elements().filter(elements => elements > 0).map(elements => elements[0])`.

And it seems the proposal `DOM.select('.foo').element()` would solve that.

If that gets implemented, I suggest for simplicity that `elements()` would always return `Stream<Array<*>>` (even for document and body) and `element()` would always return `Stream<*>`.

@Widdershin what's your perspective?

raquelxmoss:
I agree with Nick's perspective -- whenever I've used `.elements()` I've done exactly what he described. Looking forward to this being merged! Thanks very much for your work @jvanbruegge.

dom/src/MainDOMSource.ts
```diff
@@ -111,28 +116,36 @@ export class MainDOMSource implements DOMSource {
...
-    const out: DevToolEnabledSource & MemoryStream<Element> = adapt(
+    const out: DevToolEnabledSource & MemoryStream<Array<Element>> = adapt(
       output$.remember(),
     );
     out._isCycleSource = this._name;
     return out;
   }
 
+  public element(): MemoryStream<Element> {
+    return this.elements().map(arr => arr[0]);
```
Widdershin: I propose that we filter this array to remove null values.

I feel that in most use cases, when you call `.element()` you are looking to perform operations using the data in that element. In these cases, you would end up filtering out null values.

I can see a use case for when you would want to get the nulls, so that you could tell when an element had yet to render or has been removed, but I don't think that's a common usecase right now, and you could get that information from `.elements()`.

We could always make an argument that defaults to removing the nulls if we wanted to support that use case.

dom/src/DocumentDOMSource.ts
```diff
@@ -7,19 +7,23 @@ import {fromEvent} from './fromEvent';
...
+  public elements(): MemoryStream<Document[]> {
+    const out: DevToolEnabledSource & MemoryStream<Document[]> = adapt(
+      xs.of([document]),
     );
     out._isCycleSource = this._name;
     return out;
   }
 
+  public element(): MemoryStream<Document> {
+    return this.elements().map(arr => arr[0]);
```
staltz: Same comment that I made for BodyDOMSource

dom/src/MainDOMSource.ts
```diff
@@ -111,28 +116,36 @@ export class MainDOMSource implements DOMSource {
...
-    const out: DevToolEnabledSource & MemoryStream<Element> = adapt(
+    const out: DevToolEnabledSource & MemoryStream<Array<Element>> = adapt(
       output$.remember(),
     );
     out._isCycleSource = this._name;
     return out;
   }
 
+  public element(): MemoryStream<Element> {
+    return this.elements().map(arr => arr[0]).filter(Boolean).remember();
```
staltz: It's correct to reuse `elements()` here because it has the ElementFinder which we want to reuse, but before returning this remembered stream we need to set `_isCycleSource` as well as adapt() the output.

dom/src/index.ts
```diff
@@ -29,10 +29,12 @@ export {MainDOMSource} from './MainDOMSource';
  * The other option is `preventDefault` that is set to false by default.
  * If set to true, the driver will automatically call `preventDefault()` on every event.
  *
- * `DOMSource.elements()` returns a stream of the DOM element(s) matched by the
- * selectors in the DOMSource. Also, `DOMSource.select(':root').elements()`
- * returns a stream of DOM element corresponding to the root (or container) of
- * the app on the DOM.
+ * `DOMSource.elements()` returns a stream of arrays of the DOM elements matched by the selectors in the DOMSource. Also, `DOMSource.select(':root').elements()`
```
staltz: Probably nitpick, but since you'll update the commit, could keep this line less than 80 columns.

dom/src/mockDOMSource.ts
```diff
@@ -28,6 +28,10 @@ export class MockedDOMSource implements DOMSource {
     return out;
   }
 
+  public element(): any {
+    return this.elements().map((arr: any[]) => arr[0]).filter(Boolean);
```
staltz: Missing `_isCycleSource` and `adapt` here

dom/test/browser/src/dom-driver.ts
```diff
@@ -131,8 +131,8 @@ describe('DOM Driver', function() {
     let hasDisposed = false;
     let assertionOngoing = false;
     sources.DOM.select(':root').elements().drop(1).addListener({
-      next: (root: Element) => {
-        const selectEl = root.querySelector('.target') as Element;
+      next: (root: Element[]) => {
+        const selectEl = root[0].querySelector('.target') as Element;
```
staltz: Instead of introducing the array here, we could take `element` for a spin, and replace `elements()` with `element()`

dom/test/browser/src/elements.ts
```diff
@@ -0,0 +1,146 @@
...
+      DOM: makeDOMDriver(createRenderTarget()),
+    });
+
+    function isDocument(element: any): element is Document {
+      return 'body' in element && 'head' in element;
+    }
+
+    let dispose: any;
+    sources.DOM.select('document').elements().take(1).addListener({
+      next: root => {
```
staltz: Do these tests pass? I think `root` should be an array here since `elements()` was used

dom/test/browser/src/events.ts
```diff
@@ -82,8 +82,10 @@ describe('DOMSource.events()', function() {
     });
     // Make assertions
     sources.DOM.select(':root').elements().drop(1).take(1).addListener({
```
staltz: `element()` usage instead

dom/test/browser/src/events.ts
```diff
@@ -196,8 +198,10 @@ describe('DOMSource.events()', function() {
     });
 
     sources.DOM.select(':root').elements().drop(1).take(1).addListener({
```
staltz: `element()` usage instead

dom/test/browser/src/select.ts
```diff
@@ -42,10 +42,10 @@ describe('DOMSource.select()', function() {
 
     let dispose: any;
     sources.DOM.select(':root').elements().drop(1).take(1).addListener({
```
staltz: `element()` usage instead
Also applies to all other cases in this git diff

staltz:
Done. Now awaiting another review, and Travis CI build.

dom/src/DOMSource.ts
```diff
@@ -7,7 +7,10 @@ export interface EventsFnOptions {
 
 export interface DOMSource {
   select(selector: string): DOMSource;
-  elements(): MemoryStream<Document | Element | Array<Element> | string>;
+  elements(): MemoryStream<
+    Document[] | HTMLBodyElement[] | Array<Element> | string
```
jvanbruegge: I would change `Array<Element>` to `Element[]`. Reads a lot better

dom/src/index.ts
```diff
@@ -27,12 +27,16 @@ export {MainDOMSource} from './MainDOMSource';
...
+ * If set to true, the driver will automatically call `preventDefault()` on
+ * every event.
+ *
+ * `DOMSource.elements()` returns a stream of arrays of the DOM elements matched
+ * by the selectors in the DOMSource. Also,
+ * `DOMSource.select(':root').elements()` returns a stream of arrays of the DOM
+ * element corresponding to the root (or container) of the app on the DOM.
+ * Special cases are `DOMSource.select('document').elements()` that returns a
+ * Stream of documents (no arrays) and `DOMSource.select('body').elements()`
+ * that returns a Stream of HTMLBodyElements (no arrays).
```
jvanbruegge: I would mention the `element()` method.
Something like: "To get a single element there is an analog `element` function`

dom/src/DOMSource.ts
staltz: We use `Array<T>` everywhere else in the codebase, we should stay consistent in that

dom/src/index.ts
staltz: Very good point, I'll write about that.

dom/src/DOMSource.ts
jvanbruegge: Rhen we should adjust the other ones too

jvanbruegge: like `Document[]`

staltz: It's in a lot of places throughout the whole monorepo and I'm not convinced `T[]` is just "better"

staltz: Or do you mean in that line of code there? :D
I think I missed that

staltz:
Updated

staltz:
Yay! Merged! Thanks @jvanbruegge for the patience

Issue #710, opened 2017-10-08
Handle error inside `makeHtmlDriver`


desfero:
Errors should always be handled, as it may be difficult to debug when "something is not yes". 
In my case i had `undefined` in one of my children noted, and without any logging it was really hard to find what was broken.

By adding error loggin to `makeHtmlDriver` i found from what place error is thrown, in my case it was thrown form `snabbdom-to-html` package: (probably i fill also an issue there to handle `undefined` nodes)
```
TypeError: Cannot read property 'sel' of undefined
at renderToString (C:\Projects\cyclehn\node_modules\@cycle\html\node_modules\snabbdom-to-html\init.js:31:15)
```

The place where error logging should be added:
https://github.com/cyclejs/cyclejs/blob/8deedf9c91f187f77b2506a0e859e472526b7056/html/src/makeHTMLDriver.ts#L46

If any of maintainers agree with this issue i can submit a PR to fix it ;)

staltz:
Good discovery! You're right, this is a bug. We should have error reporting like cycle/dom has.

desfero:
@staltz i will submit PR today to fix it, thanks :)

PR #714, opened 2017-10-10
fix(html): error handling for makeHtmlDriver method

desfero:
ISSUES CLOSED: #710

### DIFF
--- a/html/src/makeHTMLDriver.ts
+++ b/html/src/makeHTMLDriver.ts
@@ -28,7 +28,10 @@ export interface HTMLDriverOptions {
 }
 
 export type EffectCallback = (html: string) => void;
-const noop = () => {};
+
+function reportSnabbdomError(err: any): void {
+  console.error(err);
+}
 
 export function makeHTMLDriver(
   effect: EffectCallback,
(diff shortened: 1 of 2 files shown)


html/src/makeHTMLDriver.ts
```diff
@@ -42,9 +45,8 @@ export function makeHTMLDriver(
   function htmlDriver(vnode$: Stream<VNode>, name: string): HTMLSource {
     const html$ = vnode$.map(toHTML);
     html$.addListener({
-      next: effect || noop,
```
jvanbruegge: Any reason why you removed the fallback?

desfero: Is the fallback need as effect is required?

staltz:
Hi @desfero :)
Thanks for submitting this PR. Before merging it could you notice that it's possible to add tests for this. Take a look at cycle/run tests or cycle/dom tests that stub `console.error` using `sinon` library.

html/test/index.ts
```diff
@@ -237,4 +234,22 @@ describe('HTML Driver', function() {
...
+      return {
+        html: xs.of('invalid snabbdom' as any)
+      };
+    }
+
+    run(app, {
+      html: makeHTMLDriver(() => {}),
+    });
+
+    sinon.assert.calledOnce(console.error as any);
```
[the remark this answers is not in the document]
staltz: Looks good to me! Thanks @desfero

PR #721, opened 2017-10-19
updated docs to reflect changes to preventDefault option in @cycle/do…

aronallen:
This is just an update to the docs.
I don't know how to generate the html files, any help would be much appreciated.


staltz:
Thanks @aronallen, I'll pick it up from here and adjust. Indeed our docs workflow isn't ideal at the moment for contributors.

Issue #720, opened 2017-10-18
@cycle/http 14.5.0 isolation issue (a regression from 14.2.0)


VasilioRuzanni:
Encountered an issue when upgrading `@cycle/http` to `14.5.0`. The HTTP responses don't make it to a response stream. This turned out to be isolation issue appeared after this commit: https://github.com/cyclejs/cyclejs/commit/e0dea73

**Code to reproduce the issue:**
https://jsbin.com/hinifocufu/edit?html,js,output

**Expected behavior:**
HTTP responses should work just fine through multiple levels of `isolate`ion

**Actual behavior:**
HTTP response stream never emits when `isolate`d more than 1 once with different isolation scopes.

**Versions of packages used:**

- `xstream@11.0.0`
- `@cycle/run@3.3.0`
- `@cycle/dom@18.3.0`
- `@cycle/http@14.5.0`
- `@cycle/isolate@3.1.0`

**Some insights:**
The problem is that [`arrayEqual`](https://github.com/cyclejs/cyclejs/blob/2aec2d3a0ad1db7dddc5017b23e63af574587220/http/src/isolate.ts#L4) assumes the 2 arrays of namespaces (the one in `request` and that in `isolate`d source) are exactly the same - i.e., absolutely identical, with elements going in exactly the same order, while this `request._namespaces.push(...)` [here](https://github.com/cyclejs/cyclejs/blob/2aec2d3a0ad1db7dddc5017b23e63af574587220/http/src/isolate.ts#L43) actually makes the array reverse (perhaps, because it happens in `isolateSink` and sinks are "outputs" and are collected on the way back, thus resulting in the reverse order of the `request` namespaces).

Looks like it should either:
- be a `req._namespaces.unshift(...)` [here](https://github.com/cyclejs/cyclejs/blob/2aec2d3a0ad1db7dddc5017b23e63af574587220/http/src/isolate.ts#L43); OR
- `request.namespace` should be passed in a reverse order [here](https://github.com/cyclejs/cyclejs/blob/2aec2d3a0ad1db7dddc5017b23e63af574587220/http/src/isolate.ts#L24)

The issue obviously doesn't appear when there's no `isolate` or when there's only one `isolate(...)` call (because in that case there's just one element in the `_namespaces` array).

Hopefully, I'm not missing something important, guys!

staltz:
Looks like a legit bug. Thanks for the detailed report!

whitecolor:
[...]

PR #724, opened 2017-10-26
fix(http): fix isolation regression

jvanbruegge:
ISSUES CLOSED: #720

### DIFF
--- a/http/test/browser/src/common.ts
+++ b/http/test/browser/src/common.ts
@@ -11,6 +11,7 @@ import {Observable} from 'rxjs';
 import 'rxjs/add/operator/mergeAll';
 import 'rxjs/add/operator/switchMap';
 import * as Cycle from '@cycle/rxjs-run';
+import isolate from '@cycle/isolate';
 
 export function run(uri: string) {
   describe('makeHTTPDriver', function() {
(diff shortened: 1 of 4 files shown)


staltz:
Cool. I think we should have test for this. I can do it, in case you're busy. What do you prefer?

jvanbruegge:
I dont care, I can do it in 2 hours, or you do it now

staltz:
Okay, I'll try making a test quickly here.

http/test/browser/src/common.ts
```diff
@@ -473,24 +474,24 @@ export function run(uri: string) {
...
-        .isolateSink(sources.HTTP.isolateSink(request$, 'foo'), 'something')
+      const fooInsideBarRequest$ = sources.HTTP
+        .isolateSink(sources.HTTP.isolateSink(request$, 'foo'), 'bar')
         .shareReplay();
-      const twiceScopedHTTPSource = sources.HTTP.isolateSource(
-        sources.HTTP.isolateSource(sources.HTTP, 'foo'),
-        'something',
+      const fooInsideBarHTTPSource = sources.HTTP.isolateSource(
+        sources.HTTP.isolateSource(sources.HTTP, 'bar'),
+        'foo',
```
staltz: @jvanbruegge notice that I fixed this test. It's easy to get confused, so I remind myself that

`isolateSource :: OuterSource -> InnerSource`
`isolateSink :: InnerSink -> OuterSink`

jvanbruegge: Your test would not fail on the original issue thou, the isolation was wrong because there was `foo` *somewhere* in the namespace, even of the deeper levels didnt match

So
```
isolationA: foo -> bar
isolationB: something -> foo

isolationB receives stuff from isolationA
```

staltz: Did you see the new test I added? When I revert `unshift` to `push`, both this test and the new test fail.

jvanbruegge: Do the test fail if you change it to `namespace.indexOf(scope)`?

staltz: With `namespace.indexOf(scope)`, this test would (correctly) fail on this part:

```js
      fooInsideFooHTTPSource.select().subscribe(function(response$) {
        assert(false);
        done('should not be called');
      });
```

jvanbruegge: :+1:

staltz:
Are we ready to merge? :)

jvanbruegge:
yes

Issue #725, opened 2017-10-26
Update cycle/html with latest cycle/dom changes


staltz:
In particular, new HTMLSource should implement `DOMSource.element()`

PR #726, opened 2017-10-26
fix(html): update important dependencies

staltz:
Update snabbdom. Update snabbdom-to-html. Update cycle/dom, and with it, introduce also
HTMLSource.element() in order to properly implement the DOMSource interface.

ISSUES CLOSED: #725

### DIFF
--- a/html/package.json
+++ b/html/package.json
@@ -35,22 +35,20 @@
   "types": "lib/cjs/index.d.ts",
   "dependencies": {
     "@cycle/run": "*",
-    "@cycle/dom": "18.x",
-    "snabbdom": "0.6.9",
-    "snabbdom-to-html": "3.1.1"
-  },
-  "peerDependencies": {
+    "@cycle/dom": ">=19.3.0",
+    "snabbdom": "0.7.0",
+    "snabbdom-to-html": "3.3.0",
     "xstream": "*"
   },
   "devDependencies": {
-    "@cycle/isolate": "2.1.0",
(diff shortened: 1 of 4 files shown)


html/package.json
```diff
@@ -35,22 +35,20 @@
...
-    "snabbdom-to-html": "3.1.1"
-  },
-  "peerDependencies": {
+    "@cycle/dom": ">=19.3.0",
+    "snabbdom": "0.7.0",
+    "snabbdom-to-html": "3.3.0",
     "xstream": "*"
   },
   "devDependencies": {
-    "@cycle/isolate": "2.1.0",
```
jvanbruegge: Add peer dependencies to devDependencies

staltz: Which dependencies? I removed peerDependencies because with yarn and npm5, we can use dependency `xstream: "*"` to signal that the parent package is free to specify which xstream version to use, but in the end of the day there must be some xstream version.

The case for cycle/isolate: it's only used in tests.

jvanbruegge: I meant `snabbdom` and `snabbdom-to-html`. They are needed for the tests (or at least should be), so with `peerDependencies` only they wont be installed on `make setup`

staltz: snabbdom and snabbdom-to-html are both `dependencies` in this PR

jvanbruegge: Oh, I was getting confused with the diff, sorry, all green then

PR #735, opened 2017-11-06
fix(history): fix captureClicks listener cleanup

Sinewyk:
Ensure the document event listener added when using captureClicks
is properly removed when the sink passed to the driver is disposed of

ISSUES CLOSED: #734

PR #736, opened 2017-11-07
refactor(history): tests cleanup, remove tests skip

Sinewyk:
Tests were skipped because of global state, with this cleanup
we should be able to not skip any tests ... just always cleanup everything


Sinewyk:
Rebased after https://github.com/cyclejs/cyclejs/pull/735, do what you want :+1:

jvanbruegge:
Thanks for the effort, but this was now tackled as part of #821

PR #742, opened 2017-11-14
gitignore lib/cjs and lib/es6 instead of lib

mightyiam:
This should help prevent accidental inclusion of old build artifacts in packages.

### DIFF
--- a/run/.gitignore
+++ b/run/.gitignore
@@ -8,7 +8,8 @@
 node_modules
 
 # Generated
-lib/
+lib/cjs/
+lib/es6/
 dist/
 
 # Misc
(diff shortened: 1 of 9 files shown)


jvanbruegge:
This breaks our @cycle/run build, as it involves copying files to `lib/`. I also do not see the benefit in this

run/.gitignore
```diff
@@ -8,6 +8,8 @@
 node_modules
 
 # Generated
+lib/cjs
+lib/es6
 lib/
```
mightyiam: It seems to me that @cycle/run is the only exception, where we have this backwards-compatible `lib` location. So don't ignore this.

am-a:
I believe this may have come back to haunt - as published npm rxjs-run doesn't have the latest artifacts

PR #743, opened 2017-11-14
time: lib/cjs and lib/es6 specifically in package files field

mightyiam:


mightyiam:
Actually, this looks fine to me.

Issue #744, opened 2017-11-14
Some import paths are out-dated and continue to function only because of published build artifacts


mightyiam:
Following #694, I'm still getting [errors](https://travis-ci.org/cycle-components/www/builds/302184098#L2330-L2354):

Here is one of them:

```
WARNING in ./node_modules/@cycle/run/lib/adapt.js
(Emitted value instead of an instance of Error) Cannot find source file '../../src/adapt.ts': Error: Can't resolve '../../src/adapt.ts' in '/home/travis/build/cycle-components/www/node_modules/@cycle/run/lib'
 @ ./node_modules/@cycle/dom/lib/es6/MainDOMSource.js 1:0-45
 @ ./node_modules/@cycle/dom/lib/es6/index.js
 @ ./src/index.ts
```

It surprises me that in the first line, the path `./node_modules/@cycle/run/lib/adapt.js` is shown, at all. The only version of @cycle/run that I have installed is `v4.0.0`:

```
$ npm ls | grep cycle/run               
│ ├── @cycle/run@4.0.0 deduped
│ ├── @cycle/run@4.0.0 deduped
│ ├── @cycle/run@4.0.0 deduped
├─┬ @cycle/run@4.0.0
```

In this version, the `main` and `module` fields [do not point there](https://github.com/cyclejs/cyclejs/blob/75a2481f4e7b95989a3a7a6eaa8c33107ff36779/run/package.json#L28-L29):

```
  "main": "lib/cjs/index.js",
  "module": "lib/es6/index.js",
```

So, how could this even be? Yet, when I look into what is installed, I see that this file exists in the package:

```
$ ls -l node_modules/@cycle/run/lib 
total 20
-rw-r--r-- 1 shahar shahar  207 Nov 10 11:47 adapt.d.ts
-rw-r--r-- 1 shahar shahar  722 Nov 10 11:47 adapt.js
-rw-r--r-- 1 shahar shahar  782 Nov 10 11:47 adapt.js.map
drwxr-xr-x 2 shahar shahar 4096 Nov 13 12:51 cjs
drwxr-xr-x 2 shahar shahar 4096 Nov 13 12:51 es6
```

So, I suppose that the existence of these files outside of `cjs` and `es6` is an artifact on the publisher's filesystem. #742 should help with that and while we're at it, #743.

But even if it exists, why is it used? Ahuh:

[...]

PR #749, opened 2017-11-20
import run/adapt from lib/cjs instead of lib

mightyiam:


staltz:
Hi @mightyiam, thank you for taking your time to send us this PR. We appreciate it, but have you noticed that `adapt.ts` is anyway compiled to `lib/adapt.js`? 

https://github.com/cyclejs/cyclejs/blob/d1848d163cf9a52a04b925e19cf550c60032fc78/run/package.json#L54

This was a measure we did in order to avoid the breaking change.

mightyiam:
@staltz sorry, I should have commented on this PR. Here, this is what this is about: https://github.com/cyclejs/cyclejs/issues/744#issuecomment-345654214.

Issue #751, opened 2017-11-28
RFC: Revamp isolation


jvanbruegge:
TL;DR: Use well defined keys for different types of isolation, standardize across drivers.

## Motivation

Currently it is not clear what type of isolation is used just by looking at the code. I think it would help to introduce explicit wrapper functions using Symbols to create a registry of known isolations. This also prevents accidental errors like forgetting the dot when you wanted to do sibling isolation with the DOM driver.

Also the driver can throw an error if it does not support a certain kind of isolation instead of implicitly using a different one.

## Usage

```js
import isolate, { siblingIsolation, defaultIsolation } from '@cycle/isolate';

import Child from './childComponent';

function main(sources) {
    const childSinks = isolate(Child, {
        DOM: siblingIsolation('myKey'),
        '*': defaultIsolation('myKey')
    })(sources);

    return childSinks;
}
```

## Usage in drivers

```js
import { siblingIsolationType, defaultIsolationType, isolationType } from '@cycle/isolate';

class MySource {
    public isolateSink(sink$: Stream<any>, scope: isolationType): Stream<any> {
        if(scope === siblingIsolationType || scope === defaultIsolationType) {
            //do stuff
        } else {
            throw new Error('This driver only supports sibling and default isolation');
        }
    }
    //...
}
```
For non-standard, driver-specific isolations (like the lens isolation with onionify)
```js
import { defaultIsolationType, totalIsolationType, isolationType } from '@cycle/isolate';

export const lensIsolationType = 'LENS_ISOLATION';
export funtion lensIsolation<A, B>(lens: Lens<A, B>) {
    return {
        isolationType: lensIsolationType,
        payload: lens
    };
}

class MySource {
    public isolateSink(sink$: Stream<any>, scope: isolationType | lensIsolationType): Stream<any> {
        if(scope.isolationType === totalIsolationType || scope.isolationType === defaultIsolationType) {
            //do stuff
        } else if(scope.isolationType === lensIsolationType) {
            //do other stuff
        } else {
            throw new Error('This driver only supports total, lens and default isolation');
        }
    }
    //...
}

## Implementation

```js
export const defaultIsolationType = 'DEFAULT_ISOLATION';
export const totalIsolationType = 'TOTAL_ISOLATION';
export const siblingIsolationType = 'SIBLING_ISOLATION';

//see lens isolation for factory functions
```

## Other thoughts

[...]

PR #753, opened 2017-11-28
WIP: New isolation API

jvanbruegge:
- [x] There exists an issue discussing the need for this PR (#751)


jvanbruegge:
Done so far:
- remove implicit scopes. Every scope has to be well-defined, no random scope generation any more (inside of `isolate`, you can still call it with a random scope)
- Removed tests that are not relevant any more (related to random scope generation)
- Updated tests to cope with breaking changes in the drivers (no direct string keys, but `Scope` objects

PR #755, opened 2017-11-30
feat(http): new mockHTTPDriver test utility

Sinewyk:
Use superagent-mock under the hood

ISSUES CLOSED: #567

- [x] There exists an issue discussing the need for this PR: #567 

@Widdershin thoughts ?

### DIFF
--- a/http/package.json
+++ b/http/package.json
@@ -11,12 +11,13 @@
     "@cycle/run": "*",
     "@types/superagent": "3.5.6",
     "superagent": "3.7.0",
+    "superagent-mock": "^3.6.0",
     "xstream": "*"
   },
   "devDependencies": {
+    "@cycle/isolate": "3.2",
     "@cycle/run": "4.0.0-rc.4",
     "@cycle/rxjs-run": "8.0.0-rc.4",
-    "@cycle/isolate": "3.2",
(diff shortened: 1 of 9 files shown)


http/declaration.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'superagent-mock' {
```
Sinewyk: I'm not happy myself about this file, and this was for me the hardest thing to actually make. I had trouble finding any doc on how to do it, went with somehow seeing how `@types/superagent` was made, and imitated some stuff into a `declare module 'superagent-mock' {}` scope ... It worked but no sure on how to clean this up.

staltz: That's fine, these files don't need to be perfect since they are just placeholders for the lack of proper .d.ts files. Could you @Sinewyk rename it to `superagent-mock.d.ts`?

staltz:
Overall this PR is looking good to merge, in my opinion. We're just missing some docs about this new functionality. Anyone could write that, and add it to this PR, but we anyway need it before merging and shipping.

Sinewyk:
Done. And once again, tests started parasiting each other related to the `setAdapt` stuff, so explicit over implicit and we're good to go.

Issue #687, opened 2017-09-28
Weird error stack traces that don't utilize sourcemaps


siassaj:
I'm getting weird error stack traces that don't utilize my sourcemaps;
linux chrome 59.0.3071.86
Interestingly there's no problem on my brother's osx chrome 60.0.3.3112

**Code to reproduce the issue:**
Anything causing an error after setup, as a response to a stream event. Simple example; select property on 'undefined' value.
``` js
function view(state$) {
  return state$.map(state =>{
    return div(".foo", state.doesntExist.someProperty)
  })
}
```

**Expected behavior:**
a stack track using the source map

**Actual behavior:**
a stack trace not using the source map

**Versions of packages used:**
    "@cycle/dom": "^18.3.0",
    "@cycle/run": "^3.2.0",
    "xstream": "^10.9.0"

I changed how errors are reported in @cycle/run/lib/index.js in this gist showing how I managed to get decent sourcemapped error thingo; 

https://gist.github.com/siassaj/39e83aa22bbb91364c82e6ae18c91735

Why not just throw the error again, rather than use console.error or log?

jvanbruegge:
@staltz is there a reason not to just throw the error?

staltz:
See [this comment](https://github.com/cyclejs/cyclejs/issues/158#issuecomment-161908045) and [commit](https://github.com/cyclejs/cyclejs/commit/ceb58a5f93e88fdb422b8ef743e484fdfc3c455e) for the historical context how this came to be.

I believe the reason why that commit used console.error instead of throwing was because at that time we used RxJS v4, which had its own [trampoline scheduler](https://github.com/Reactive-Extensions/RxJS/blob/6770d24307ed4baaecdbb21cfc178c13e459f42f/src/core/scheduledobserver.js#L23). If we had thrown the error, it would just end being swallowed, so we had to report the error in the observer.error somehow.

I'm open to trying the rethrow approach, but we need to experiment with this carefully, because the logToConsoleError is in cycle/run, simply taking a peek into the stream graph. The errors also go into sinkProxies and into drivers, and drivers may decide how to handle errors in their own way.

If our goal is to simply support proper stack traces, I think we can ask ourselves why exactly is it that chrome 59 doesn't utilize sourcemaps for `console.log(err.stack)` while chrome 60 does. And how does Firefox and Safari behave?

siassaj:
[...]

PR #756, opened 2017-12-07
fix(run): make exception reporting less opinionated

staltz:
Error/exception reporting no longer expects there to be a stack trace on the error object. Any
error, even if it was `throw 'error-as-string'` will be reported as is.

ISSUES CLOSED: #687

### DIFF
--- a/run/src/index.ts
+++ b/run/src/index.ts
@@ -29,15 +29,6 @@ export {
 
 const scheduleMicrotask = microtask();
 
-function logToConsoleError(err: any) {
-  const target = err.stack || err;
-  if (console && console.error) {
-    console.error(target);
-  } else if (console && console.log) {
-    console.log(target);
-  }
-}
-
 function makeSinkProxies<So extends Sources, Si extends Sinks>(
   drivers: Drivers<So, Si>,
 ): SinkProxies<Si> {
@@ -133,7 +124,7 @@ function replicateMany<So extends Sources, Si extends Sinks>(
     };
     const error = (err: any) => {
       scheduleMicrotask(() => {
-        logToConsoleError(err);
+        (console.error || console.log)(err);
         listener._e(err);
       });
     };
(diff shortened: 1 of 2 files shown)


run/test/index.ts
```diff
@@ -407,12 +407,13 @@ describe('run', function() {
...
-            url: 'http://jsonplaceholder.typicode.com/users/1',
-          })
-        : xs.never();
+      const request$ =
+        num === 1
+          ? xs.of({
+              category: 'cat',
+              url: 'http://jsonplaceholder.typicode.com/users/1',
+            })
+          : xs.never();
```
staltz: This was done by the automatic `prettier` runner

run/test/index.ts
```diff
@@ -421,16 +422,19 @@ describe('run', function() {
...
+        .map(i => {
+          if (i % 2 === 1) {
+            return child(sources, i);
+          } else {
+            return {
+              HTTP: xs.empty(),
+              DOM: xs.of(''),
+            };
+          }
+        });
```
staltz: This was done by the automatic `prettier` runner

run/test/index.ts
```diff
@@ -440,16 +444,19 @@ describe('run', function() {
...
```
staltz: This was done by the automatic `prettier` runner

run/src/index.ts
```diff
@@ -133,7 +124,7 @@ function replicateMany<So extends Sources, Si extends Sinks>(
     };
     const error = (err: any) => {
       scheduleMicrotask(() => {
-        logToConsoleError(err);
+        (console.error || console.log)(err);
```
jvanbruegge: Why not throw the error directly?

staltz: To avoid error swallowing: https://github.com/cyclejs/cyclejs/issues/687#issuecomment-333814900

jvanbruegge:
LGTM

Issue #758, opened 2017-12-15
Sibling isolation is not working when using the className prop


jvanbruegge:
**Code to reproduce the issue:**
```js
function Child(sources) {
    sources.DOM.select('.btn').events('click').addListener({ next: console.log });
    return {
        DOM: xs.of(div({ props: { className: 'myClass' }}, [button('.btn', ['Button'])]))
    };
}
function Parent(sources) {
    const sinks = isolate(Child, '.foo')(sources);
    return sinks;
}
function App(sources) {
    return isolate(Parent, 'bar')(sources);
}
```

**Expected behavior:**
On click of the button I get a console log

**Actual behavior:**
No console log

**Reason:**
The className prop overwrites the css class that the dom sibling isolation is adding for the sibling isolation, so it wont match the expected set of css classes.

**Versions of packages used:**
DOM: 20.1.0
isolate: 3.2.0

Issue #747, opened 2017-11-17
`remove` animation on components breaks DOM isolation


atomrc:
**Code to reproduce the issue:** 

Live demo : https://www.webpackbin.com/bins/-Kz9T7jtEIMlufYbmkyY

```javascript
import { run } from "@cycle/run";
import { makeDOMDriver, div, button } from "@cycle/dom";
import isolate from "@cycle/isolate";
import onionify, { makeCollection } from "cycle-onionify";
import xs from "xstream";

function Child(sources) {
  const remove$ = sources.DOM.select(".click")
    .events("click")
    .mapTo(state => ({ show: false }))
  
  const style = {
    transition: "transform 1s",
    /** the remove handler seems to break isolation **/
    remove: {
      transform: "translateY(100%)"
    }
    /** END **/
  }
  
  return {
    DOM: xs.of(button(".click", { style: style }, "remove")),
    onion: remove$
  }
}

function main(sources) {
  
  const initialReducer$ =  xs.of(() => ({ show: true }));
  
  const child = isolate(Child, { onion: null })(sources);
  
  const showChild$ = sources.DOM.select(".click")
    .events("click")
    .mapTo(state => ({ show: true }))
  
  return {
    DOM: xs.combine(sources.onion.state$, child.DOM)
      .map(([state, child]) => div(
        [button(".click", "show"), (state.show ? child: null)])
      ),
    onion: xs.merge(initialReducer$, showChild$, child.onion)
  }
}

run(onionify(main), {
  DOM: makeDOMDriver("#main")
});
```

**Expected behavior:**
The remove button, when clicked, should disappear and only the `show` button should be visible.

**Actual behavior:**
The remove button, when clicked, disappears but another remove button is also added

**Versions of packages used:**
    cycle/dom    20.1.0
    cycle/isolate   3.2.0
    cycle/run    4.0.0
    cycle-onionify    4.0.0

staltz:
Hi @atomrc  :)
Thanks for letting us know about this. Yes, it seems related to isolation. When used like `isolate(Child, {onion: null})`, it means there is a randomly generated isolation scope given to the DOM channel.

I'm investigating what is the root cause of this, but for now, notice that if you wrap the child's button:

```js
DOM: xs.of(div([button('.click', {style:style}, 'remove')])),
```

Then the issue goes away. So it seems like a logic bug in the DOM isolation layers **OR** an issue with documentation or intended usage of this API. I'll investigate.

[...]

Issue #700, opened 2017-10-03
Polyfills


Sinewyk:
I've seen packages saying

> _insert lib here_ depends on the collection types Map and Set. If you support older browsers and devices which may not yet provide these natively (e.g. IE < 11), consider including a global polyfill in your bundled application, such as core-js or babel-polyfill.

But cyclejs does not actually give you a choice and bundles everything by default. I've just started doing a small application, and I just [webpack-bundler-analyzed it](https://github.com/webpack-contrib/webpack-bundle-analyzer) ... and this is what I got

![cyclejs_polyfills](https://user-images.githubusercontent.com/1826366/31151478-f5f278ca-a897-11e7-898f-a8d6572afaa2.jpg)

There's xstream, snabbdom(-jsx too), the dom driver, everything looks normal. Then on the right (everything circled in red) I see actually a tons of polyfill I didn't ask for.

Anyway we could work on pushing them out of the core so that users can decide whether to include them or not ?

jvanbruegge:
i think this would be a good thing, we could additionally publish a library that bundles all required polyfills, so the users that do want polyfills can simply use `import '@cycle/polyfills'`.

On a side note, `snabbdom-jsx` js very buggy, especially with svg, you might want to look into `snabbdom-pragma`

geovanisouza92:
Why there is two xstream?

@Sinewyk just for curiosity, how did you created this app? Using create-cycle-app or manually?

whitecolor:
Currently one may use WebpackIgnorePlugin or `alias` to exclude from the bundle. But I'm too would exclude polyfills from modules. And just would give the instructions, how to include them if needed.

Sinewyk:
Wow, I thought it was just the xstream cycle/run adapter. But no, well spotted.

[@cycle/run has a hard dependency on xstream 10.x](https://github.com/cyclejs/cyclejs/blob/41d7759302d474779ce8888a3b9cc09924681966/run/package.json#L33) , I pulled xstream 11 when installing recently, and yarn locked it down correctly. So I'm pulling both 10.9 and 11.0 into my bundle.

That sucks too. It must be the same for almost anyone using a bundler with current package versions.

staltz:
> @cycle/run has a hard dependency on xstream 10.x , I pulled xstream 11 when installing recently, and yarn locked it down correctly. So I'm pulling both 10.9 and 11.0 into my bundle.

[...]

Issue #587, opened 2017-04-25
_delegators keeps reference to past EventDelegator instances


aronallen:
```typescript
function component (sources) {
  const vdom$ = sources
    .select('div')
    .events('click')
    .startWith(null)
    .map(() => h('div', {}, []));
  return {
    DOM: vdom$
  }
}

function main (sources) {
  const components$ = periodic(1000)
       .map(() => Array(4).fill(null).map(() => isolate(component)(sources)))
  components$
    .map(components => components.map(({ DOM }) => DOM))
    .map(DOM => combineArray(Array, DOM))
    .switch()
    .map(children => h('div', children))
  return {
    DOM: 
  }
}
```

**Expected behavior:**
When the isolated module is no longer in use, I would expect the event delegators to be dereferenced.

**Actual behavior:**
The event delegators remain referenced, causing a memory leak.

**Versions of packages used:**

"@cycle/dom": "^17.1.0",
"@cycle/isolate": "^2.1.0",
"@cycle/most-run": "^7.1.0"
"most": "^1.2.2",

michalvankodev:
Hey @aronallen ! Have you also tried this with `@cycle/run` and xstream?

aronallen:
No I haven't tried that yet, but looking through the code I couldn't find
anywhere where elements where removed from _delegators.

aronallen:
I think a solution could be to pass a reference to _delegators to each EventDelegator upon initialization, and in EventDelegator stop method, to remove the reference from there.

staltz:
Thanks Aron, seems legit.

jvanbruegge:
This is now fixed in `@cycle/dom` v21 :tada:

PR #759, opened 2017-12-15
Fix sibling isolation implementation in DOM driver

jvanbruegge:
This PR will remove the use of CSS classes for sibling isolation.
Currently I just added a test that fails in the current implementation

ISSUES FIXED: #758, #587, #700, #243, #545
NEEDS TEST: #333
WORK IN PROGRESS: #775, #747

# What the *fish* I did here:
## Definitions:
- A **namespace** is an Array of scopes.
- A **scope** is an object with a type and a value
- A scope **type** is either total, sibling or selector
- A scope **value** can be anything serializable and is used to distinguish the scopes

## Algorithm
- Work is divided into two main parts: Isolation (IsolateModule) and event handling (EventDelegator)
    * All event handling code was removed from the MainDOMSource
- Isolate module expects **namespace** on vnode on create (only for isolation roots), saves element in tree.
    * Top level is `[]` the root namespace without isolation.
    * Traversing the tree and collecting the **scope values** gives back **namespace**
    * Using a special Symbol you can get the Element of that isolation scope (to avoid collision with **scope value**)
    * Namespace can be queried by looking for parent of Element in namespace tree
- Event delegator adds a virtual listener in namespace tree
    * total and sibling are saved as nodes in the tree, selector is saved extra
    * has actual listeners at the root element for bubbling events
    * non bubbling events are saved in a Map of Maps (eventType -> Element -> Listener)
    * resets listener on element updates (avoids recreation)
    * simulates capture phase with inverse bubbling path
    * simulates bubble phase
## Other stuff that also happened
- No polyfills any more
- More ES6 features used (Map, Set and Symbol vs. just Map)
- No observable change to the vnodes (no additional css classes etc)
## Why are the issues Fixed
- #758: No css classes added to the elements (verified by test)
- #587: Just one delegator at any time in the program
- #700: Just deleted the polyfills
- #243: Symbol usedTree to avoid name collision

## What still needs work
- [ ] ~~#540: Add test for manipulation of head/body/html~~ (let's do it separately)
- [x] #775: Check if rxjs and most typings are up-to-date
- [x] Check why **sometimes** `DOMSource.events() should catch non-bubbling events in a DocumentFragment with useCapture` and `isolation should handle events when child is removed and re-added` fail in older Chrome Browsers
- [x] examples/advanced/many works correctly
- [ ] Document the breaking changes
- [ ] #747: Add test for remove animatios
- [ ] #333: Add test to check non-bubbling listener cleanup

<hr>

- [x] There exists an issue discussing the need for this PR #758


staltz:
Cool. Do you intend to do the fix in `h`?

jvanbruegge:
no, in the isolation implementation

jvanbruegge:
im currently on it

dom/src/DOMSource.ts
```diff
@@ -14,6 +15,7 @@ export interface DOMSource {
   events<K extends keyof HTMLElementEventMap>(
     eventType: K,
     options?: EventsFnOptions,
+    bubbles?: boolean,
```
staltz: Shouldn't we put `bubbles` inside `options`?

jvanbruegge: I put it extra because it is only used at event listener creation. The options are used later to match individual events or to simulate capture

staltz:
Jan, would you describe your this PR as "global event delegation"? I'm not sure if you're familiar with React's event delegation, but basically it adds event listeners to the root container element only, and then simulates everything else. It's called synthetic event system. Is this PR the same thing?

There are pros and cons of having a synthetic event listeners. I heard it doesn't play well with Shadow DOM (web components) and some custom events, but it's also important to ask ourselves if those specific downsides will affect the use cases of Cycle.js with web components. Maybe it will, maybe it won't.

staltz:
And, what is the primary problem that synthetic event listeners solve here, is it the isolation, or overall performance?

staltz:
Re: the downsides

> I heard it doesn't play well with Shadow DOM (web components) and some custom events

We could have some new tests for use cases like
- Using a web component inside my Cycle app still allows me to listen to the WC's events
- DOMSource can pick events of a custom event type like any other event type

jvanbruegge:
Yes, all bubbling events are only handled by the root element, the capture and bubble phase is simulated.

Performance was a minor motivation, but it's more about simlicity and predictability. For example the memory leak we had before (#587) and the complex logic in the MainDomSource.

AFAICT it should work well with web components, but yeah, a few additional tests are neccesary

jvanbruegge:
> DOMSource can pick events of a custom event type like any other event type

That should work even better than before, because now you can also use custom non-bubbing events (because you can tell `events` that an event does not bubble) not sure how that worked before

staltz:
Yeah, I'm just mentioning these issues because we need to be sure we understand the possible downsides and how to solve them. On Monday and Tuesday I might be able to add some additional tests for this kind of stuff.

dom/src/DOMSource.ts
staltz: Still, couldn't it be an option? I'm wondering whether this is a public or private API. It's used in the tests, so I'd guess a public API, which is why it should be an option.

dom/src/utils.ts
```diff
@@ -46,18 +46,51 @@ export function getValidNode(
...
+  }
+  for (let i = 0; i < a.length; i++) {
+    if (a[i].type !== b[i].type || a[i].scope !== b[i].scope) {
+      return false;
+    }
+  }
+  return true;
+}
+
+export function getTotalIsolatedScope(namespace: Array<Scope>): Array<Scope> {
```
staltz: This function is in a hot path: EventDelegator.onEvent -> EventDelegator.buildPath -> getTotalIsolatedScope, so we should avoid slice+splice which are not the most efficient way of writing this function. We can build the result array by construction, instead of cloning then removing.

dom/src/EventDelegator.ts
```diff
@@ -53,116 +82,345 @@ function indexOf(arr: Array<Destination>, searchId: number): number {
...
-      if (matchesSelector(ev.target as Element, dest.selector)) {
-        dest.subject._n(ev);
-      }
-    }
+  private onEvent(
+    type: string,
+    event: Event,
+    passive: boolean,
+    bubbles = true,
+  ): void {
```
staltz: I'm worried about this function. It calls
- getTotalIsolatedScope (does slice + splice)
- buildPath: allocates an array and has a while loop
- slice and reverse
- bubble twice: which allocates a `Set` and has 3 nested loops: while-for-for

And it's on hot code path, so imagine a scroll event or a mousemove event on the DOM calling `onEvent` frequently.

staltz:
I published `@cycle/dom@21.0.0-experimental` which has this PR. Then I tried using that in `examples/advanced/many` and the app startup didn't work, I got the error

```
Uncaught TypeError: Cannot read property 'addEventListener' of undefined
    at Object.start (main.js:744)
    at Object._start (main.js:9761)
    at Stream._add (main.js:10584)
    at Stream.addListener (main.js:10645)
    at Stream.subscribe (main.js:10663)
    at EventDelegator.setupDOMListener (main.js:290)
    at EventDelegator.addEventListener (main.js:188)
    at MainDOMSource.events (main.js:627)
    at intent (main.js:11570)
    at List (main.js:11651)
```

Maybe `.select( ).events( )` called `setupDOMListener` before `rootElement$` emits (which should update this.origin in EventDelegator)

dom/src/EventDelegator.ts
jvanbruegge: Should be fixed now

jvanbruegge:
@staltz the `Cannot read property 'addEventListener' of undefined` should be fixed too

jvanbruegge:
CI build breaks because of history driver, not related to DOM

dom/src/EventDelegator.ts
staltz: Thanks! I'll take a look

staltz:
Jan, I added some commits for small refactors.

Could you update the list of things to be done in the first comment in this PR? E.g. what about migration docs, tests, performance considerations, etc? What's missing? Then either you or me can work on it.

jvanbruegge:
Done, basicly just adding two tests and check the typscript typings

jvanbruegge:
We also have to decide if we want `quick-lru` as additional cache
Pro:
- Makes namespace lookup cheaper (especially for events that fire often like `mousemove`)

Con:
- more code to pull in

staltz:
`quick-lru` addition is something to-do, but not for this PR. We can do it later. Now the priority is just to get the last bits solved, then release as RC. After it's released as official, we can start adding the `quick-lru` or similar perf improvements.

jvanbruegge:
I took a look at the test and it's always the same tests. Out of 4 CI logs:
| Count | Browser | Test Name |
|--------|-----------|---------------|
| 4 | Chrome 49.0 | DOMSource.events() should catch non-bubbling events in a DocumentFragment with useCapture |
| 2 | Chrome 30.0 | isolation should handle events when child is removed and re-added |

jvanbruegge:
@staltz passing in all browsers on Browserstack, ready for final review and merge!

staltz:
This PR to me is ready to merge, I wouldn't make those two tests a blocker for merging. Just needs one final thing, the documentation of the breaking changes, just need to write it in `./docs/content/releases.md` under the title "Synthetic DOM events".

jvanbruegge:
ok, i will mark the test as to be skipped and add thr breaking changes

jvanbruegge:
Ready to merge now :tada:

staltz:
I tried from mobile but couldn't. Can you merge? :)

staltz:
I'll release v21 and tweet from the official account : )

jvanbruegge:
As you already saw, I also published the article

atomrc:
Wouhou. Congrats guys. Will try this out soon for sure :+1:

PR #761, opened 2017-12-28
chore(time): use import syntax to support rollup

Widdershin:


time/sorted-immutable-list.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module 'sorted-immutable-list';
```
jvanbruegge: I would put them all into one `custom-typings.d.ts`

time/src/assert-equal.ts
```diff
@@ -1,6 +1,6 @@
 import xs, {Stream} from 'xstream';
 import {deepEqual} from 'assert';
-const variableDiff = require('variable-diff');
+import * as variableDiff from 'variable-diff';
```
jvanbruegge: Not sure if this is the same. Let's see if the tests pass

time/src/scheduler.ts
```diff
@@ -1,5 +1,5 @@
 import {Listener} from 'xstream';
-const makeAccumulator = require('sorted-immutable-list').default;
+import {makeAccumulator} from 'sorted-immutable-list';
```
jvanbruegge: Shouldnt this be `import makeAccumulator from ...`, (the default import)?

time/sorted-immutable-list.d.ts
Widdershin: I was thinking about that, but wasn't sure if it would work, but I agree and have done it now. 😄

time/src/assert-equal.ts
Widdershin: Yeah, this should be the same, and the tests pass locally.

```js
const value = require('foo');
import * as value from 'foo';

const value = require('foo').default;
import value from 'foo';

const {value} = require('foo');
import {value} from 'foo';
```

The only reason why this doesn't feel right is because for the longest time both `TypeScript` and `babel` shipped a shim that would allow you to `import value from 'foo'` and have it work both as `require('foo')` and `require('foo').default`, depending on whether the module had `__esmodule` set.

However, they both correctly follow the import spec now, so all is well in the world.

(Not super related to this PR, just thought it might be interesting for anyone reading).

In any case, I agree, let's see what CI says!

time/src/scheduler.ts
Widdershin: Haha yes, saw that after running the tests!

Widdershin:
Will merge and deploy once we've seen that the time tests pass in CI. Thanks for the über quick review @jvanbruegge!

Widdershin:
Only failure was:

> Mobile Safari 8.0.0 (iOS 8.4.0) historyDriver - xstream should allow going back/forwards with `go`, `goBack`, `goForward` FAILED
	AssertionError: '/test' === '/other' (/tmp/karma-typescript-bundle-6830yMisz03aF5d4.js:1829)
Mobile Safari 8.0.0 (iOS 8.4.0): Executed 11 of 19 (1 FAILED) (skipped 8) (1.388 secs / NaN secs)

All of the time tests passed. We're good to go.

Widdershin:
Released `@cycle/time@0.11.0` with these fixes.

PR #768, opened 2017-12-29
Update yarn-x-for-all-packages.sh

joewoodhouse:


jvanbruegge:
Not relevant any more

PR #773, opened 2018-01-08
fix(time): fix throttle immediately completing with sync producer

### LINKED ISSUE #770: @cycle/time 0.12.0 changes operator behaviour in breaking/potentially surprising way
```javascript

const { mockTimeSource } = require('@cycle/time')
const xs = require('xstream').default

const time = mockTimeSource({ interval: 1000 })

xs.from([0, 1, 2, 3, 4, 5])
.compose(time.throttle(1000))
.subscribe({
  next: (value) => {
    console.log('The Stream gave me a value: ', value);
  },
  error: (err) => {
    console.error('The Stream gave me an error: ', err);
  },
  complete: () => {
    console.log('The Stream told me it is done.');
  },
})

time.run()
```

Package.json being
```
{
  "name": "cycletimethrottletest",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "@cycle/time": "0.12.0",
    "xstream": "^11.1.0"
  }
}
```

**Expected behavior:**
On @cycle/time v0.11.0:
```
The Stream gave me a value:  0
The Stream told me it is done.
```

**Actual behavior:**
On @cycle/time v0.12.0:
```
The Stream told me it is done.
```

cycle/time: 0.11.0 -> 0.12.0
xstream: 11.1.0

Is this an intentional change? I was alerted to this change on upgrading to get the latest and greatest changes in cycle/time and was surprised by 99 of our 300 or so unit tests suddenly breaking. :)

I suspect that https://github.com/cyclejs/cyclejs/commit/fbee20cb16859b7c9de3c2526da6aa74b5bac200#diff-07cc694b9b3fc636710fa08b6922c42b
is causing the change here, but I'm not sure if it's an intentional change (and that my unit tests were passing before, but only by accident/exploiting some now invalid behaviour), or if this is an unintended consequence.

Essentially now that stream completion is propagated immediately instead of waiting, it beats out any events that are about to be pushed through. Maybe it should check to see if there are any extant, waiting events that have been scheduled to essentially immediately be next'd (don't ask me how hard 'essentially immediately' is to measure...!) but haven't yet.

You can verify this behaviour with this:
```javascript
xs.merge(
  xs.from([0, 1, 2, 3, 4, 5]),
  xs.never()
)
.compose(time.throttle(1000))
.subscribe({
  next: (value) => {
    console.log('The Stream gave me a value: ', value);
  },
  error: (err) => {
    console.error('The Stream gave me an error: ', err);
  },
  complete: () => {
    console.log('The Stream told me it is done.');
  },
})
```
on v0.12.0, which will output:
```
The Stream gave me a value:  0
```

This originally was causing very weird errors like in my test suite like:
```
Expected

-----------------------------({"foo":"bar"}|)

Got

#

Failed because:

 * Length of actual and expected differs
 * Expected type next at time 0 but got error
 * Unexpected error occurred

Unexpected error:
 EmptyError: no elements in sequence
    at new EmptyError (/-----/node_modules/rxjs/src/util/EmptyError.ts:13:22)
    at FirstSubscriber._complete (/-----/node_modules/rxjs/src/operators/first.ts:163:25)
    at FirstSubscriber.Subscriber.complete (/-----/node_modules/rxjs/src/Subscriber.ts:122:12)
```

Oddly enough turning my input streams into infinite versions fixed the problem, which having traced it back to this reproducible example seems to be due to streams completing before they push their values through, whereas infinite streams will never complete.

Widdershin:
v0.12.0 introduced a regression where applying throttle to synchronous
source (like xs.of or xs.from) would cause the resulting stream to
immediately complete, before firing an event.

This was not a documented behaviour, but @cycle/time consistently
processes next events prior to completion events if they occur in the
same frame.

I have added a unit test to prevent any regressions to this behaviour in
the future.

Thanks to @willheslam for reporting this issue.

This fixes #770.

### DIFF
--- a/time/src/assert-equal.ts
+++ b/time/src/assert-equal.ts
@@ -30,7 +30,9 @@ function checkEqual(
 
     if (actual.type !== expected.type) {
       failReasons.push(
-        `Expected type ${expected.type} at time ${actual.time} but got ${actual.type}`,
+        `Expected type ${expected.type} at time ${actual.time} but got ${
+          actual.type
+        }`,
       );
     }
 
@@ -41,7 +43,9 @@ function checkEqual(
 
       if (!rightTime) {
         failReasons.push(
-          `Expected stream to complete at ${expected.time} but completed at ${actual.time}`,
+          `Expected stream to complete at ${expected.time} but completed at ${
+            actual.time
+          }`,
         );
       }
     }
@@ -71,7 +75,9 @@ function checkEqual(
 
       if (!rightTime || !rightValue) {
         const errorMessage = [
-          `Expected value at time ${expected.time} but got different value at ${actual.time}\n`,
+          `Expected value at time ${expected.time} but got different value at ${
+            actual.time
+          }\n`,
         ];
 
         if (usingCustomComparator) {
--- a/time/src/throttle.ts
+++ b/time/src/throttle.ts
@@ -31,7 +31,7 @@ function makeThrottleListener<T>(
     },
 
     complete() {
-      listener.complete();
+      schedule.complete(listener, currentTime());
     },
   };
 }
--- a/time/test/time-driver.ts
+++ b/time/test/time-driver.ts
@@ -1,4 +1,3 @@
-
 import * as assert from 'assert';
 import {timeDriver, TimeSource, Operator} from '../';
 import xs, {Stream} from 'xstream';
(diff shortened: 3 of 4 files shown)


willheslam:
Thanks for working on this fix @Widdershin, I really appreciate it! :)

Should this also apply to debounce?
https://github.com/cyclejs/cyclejs/blob/4b7a2141cecb700ac9eaa1204b8f5ea56ac854a3/time/src/debounce.ts#L33
(it already applies to delay)

(I don't have enough experience with throttleAnimation to know if it should apply there either: https://github.com/cyclejs/cyclejs/blob/4b7a2141cecb700ac9eaa1204b8f5ea56ac854a3/time/src/throttle-animation.ts#L40 )

Regarding the comparison to delay: delay delays both next, error and complete.
Can this bug also manifest with errors being emitted instead of streams completed?
I had hoped to help contribute some unit tests for this sort of thing this weekend but I ended up having a bad cold - here's an example of what I mean:
```javascript
const time = mockTimeSource({ interval: 100 })

xs.from([xs.of(1), xs.of(2), xs.of(3), xs.throw('Error!')])
.flatten()
.compose(time.throttle(1000))
.subscribe({
  next: (value) => {
    console.log('The Stream gave me a value: ', value);
  },
  error: (err) => {
    console.error('The Stream gave me an error: ', err);
  },
  complete: () => {
    console.log('The Stream told me it is done.');
  },
})

time.run()
```

I'm expecting this I suppose:
```
The Stream gave me a value:  1
The Stream gave me an error:  Error!
```

But with the immediate error propagation I get:
```
The Stream gave me an error:  Error!
```

PR #774, opened 2018-01-15
Add required modules for babel/jsx to npm install.

bloodyKnuckles:
I needed to also install babel-cli and babel-preset-es2015 to run a basic jsx example. In fact I also need babelify since I was using browserify, but didn't add that to the example since the bundler is not specified.


staltz:
Hi :)
Thanks for submitting this PR, superficially it looks correct. Before merging this PR I think we need someone else to confirm that this is a normal requirement when using Babel.

jvanbruegge:
It looks ok, but I wouldnt advertise snabbdom-jsx in the docs. It's buggy and unmaintained. Snabbdom-pragma works a lot better.

PR #777, opened 2018-01-17
 docs(suggestion): use template string for "Send Messages from main" section

AgtLucas:

**This is just a suggestion.**

In the section ["Send messages from main"](https://cycle.js.org/getting-started.html#getting-started-coding-send-messages-from-main), instead of:

```javascript
'' + i + ' seconds elapsed'
```

I think is better to use template string, so we can actually match the documentation that says: `The stream emits Virtual DOM <h1> elements displaying ${i} seconds elapsed changing over time every second, where ${i} is replaced by 0, 1, 2, etc.`

What do you think?

### DIFF
--- a/docs/getting-started.html
+++ b/docs/getting-started.html
@@ -132,7 +132,7 @@ We have filled the `main()` function with some code: returns an object `sinks` w
 function main() {
   const sinks = {
     DOM: xs.periodic(1000).map(i =>
-      h1('' + i + ' seconds elapsed')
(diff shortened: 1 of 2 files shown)


staltz:
Hi, thank you for taking your time to send us this PR. We appreciate it, however I think the original reason why we used basic string concatenation was the syntax highlighter in HTML didn't support it.

AgtLucas:
@staltz No problem! :)

PR #781, opened 2018-02-11
docs(examples): fix import path in animation example

allonhadaya:


staltz:
Thanks!

PR #784, opened 2018-02-13
Fixed small error in BMI calculator basic example

### LINKED ISSUE #783: In bmi documentation, missing attrs field in slider input
There is a small error in the documentation for the final listing of Body mass calculator in Basic Examples. Looks like the input functions are passing in an attribute map without placing them into a property map with key attrs.

These:
input('.weight', {type: 'range', min: 40, max: 140, value: weight})
input('.height', {type: 'range', min: 140, max: 210, value: height})

Should be:
input('.weight', {attrs: {type: 'range', min: 40, max: 140, value: weight}})
input('.height', {attrs: {type: 'range', min: 140, max: 210, value: height}})

alexdennis:
In the final listing of the BMI calculator basic example there is a small error with the input declaration. It's missing the attr key in the config object.

Fixes #783

### DIFF
--- a/docs/basic-examples.html
+++ b/docs/basic-examples.html
@@ -515,11 +515,11 @@ function main(sources) {
     div([
       div([
         'Weight ' + weight + 'kg',
-        input('.weight', {type: 'range', min: 40, max: 140, value: weight})
+        input('.weight', {attrs: {type: 'range', min: 40, max: 140, value: weight}})
       ]),
       div([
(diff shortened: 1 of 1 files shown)


staltz:
Hi @alexdennis  :)
Thanks for submitting this PR, we appreciate any kind of desire to help. Note that the html files are not meant to be edited directly because they are generated from the markdown files in `docs/content`. I'll do these markdown changes and generate the html, no worries

Issue #785, opened 2018-02-19
MainDOMSource.element calls xstream specific methods on adapted streams


siliconbrain:
**Code to reproduce the issue:**
```
const {makeDOMDriver} = require('@cycle/dom');
const {run} = require('@cycle/rxjs-run');
run(
  function({DOM}) {
    DOM.select('#some-element').element()  // --> 'remember' is not a function
  },
  {
    DOM: makeDOMDriver('.container')
  }
)
```

**Expected behavior:**
Returns stream of selected element.

**Actual behavior:**
An error is thrown. The problem is caused by [this line](https://github.com/cyclejs/cyclejs/blob/4bdb8ddf9e890a4771a3f6d26cf79708d211d2f0/dom/src/MainDOMSource.ts#L137) reusing `MainDOMSource.elements`'s return value which has been adapted to the target library (in this case, RxJS).

**Versions of packages used:**
`@cycle/dom`: 20.1.0
`@cycle/rxjs-run`: 8.0.0

staltz:
Hi :)
Thanks for letting us know. Looks legitimate. I'll try to fix it asap.

staltz:
DOM v20.2.0 released

PR #786, opened 2018-02-22
fix(dom): fix .elements() support for rxjs and most

staltz:
ISSUES CLOSED: #785


staltz:
It's now green in Travis. Good to go?

jvanbruegge:
yes, changing labels is just a pain on mobile, thats why i just approved it

PR #788, opened 2018-02-27
chore(META): replace bump-cli with simple bump script

### LINKED ISSUE #737: Replace `bump-cli` tool
This concerns only our monorepo tools, but is still important. I just experienced this problem when releasing cycle/dom `110.0.0` (lol).

Try to use `bump` CLI to update a major number from `19.0.0`.

**Expected behavior:**

`20.0.0`

**Actual behavior:**

`110.0.0`

Widdershin:
Closes #737.

Unlike bump-cli, bump.js can bump from 19.0.0 to 20.0.0, rather than to
110.0.0.

### DIFF
--- a/.scripts/bump.js
+++ b/.scripts/bump.js
@@ -1,0 +1,32 @@
+#!/usr/bin/env node
+const path = require('path');
+const fs = require('fs');
+
+const packagePath = process.argv[2];
+const bumpType = process.argv[3];
+
+if (!packagePath || packagePath === '') {
+  console.log("Please provide a relative package.json path as the first argument");
+  process.exit(1);
+}
+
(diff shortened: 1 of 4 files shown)


Widdershin:
@staltz the only thing I'm unsure about here is if I need to support `-rc2` style versions, or if we're doing those by hand.

Wouldn't be hard to do, just doesn't set the `Makefile` is currently set up for any prerelease automation, so I wasn't sure if it was needed.

staltz:
Thanks a lot for this PR @Widdershin ! I like the simple implementation, because then it brings confidence that it'll work.

> the only thing I'm unsure about here is if I need to support -rc2 style versions, or if we're doing those by hand.

No need to support `-rc`, so far I've been doing those types of releases by hand.

PR #789, opened 2018-02-27
docs(time): migrate README to api docs

Widdershin:
In relation to #635.


Widdershin:
There's an issue with this right now, it appears the markdown parser is choking on the comments that markdown-doctest uses.

```
/Users/nickj/Projects/cyclejs/node_modules/dox/lib/dox.js:30
  js = js.replace(/\r\n/gm, '\n');
         ^

TypeError: Cannot read property 'replace' of undefined
  at Object.exports.parseComments (/Users/nickj/Projects/cyclejs/node_modules/dox/lib/dox.js:30:10)
  at ReadFileContext.callback (/Users/nickj/Projects/cyclejs/node_modules/markdox/lib/markdox.js:152:23)
  at FSReqWrap.readFileAfterOpen [as oncomplete] (fs.js:366:13)
```

Guess we can see about fixing dox, assuming it's a problem there.

staltz:
Good PR overall. About those errors with markdown-doctest, I think I've seen it before too. Fixing dox would be one approach, but we could check if they have an update and whether we are using an older version.

Widdershin:
Yep, I'll try updating it first, after verifying dox is in fact the issue.

Just ran out of time on this yesterday so I figured I would push up what I had 😄

PR #790, opened 2018-03-01
docs(META): use ES6 syntax in getting-started

BlackYoup:
The text above that piece of code is using ES6 syntax to reference `i`

I didn't mind making an issue for this small change. Apologies if I should have done it.


jvanbruegge:
The problem is that the syntax highlighter does not support it

Issue #540, opened 2017-03-05
Request: DOMDriver to accept `Stream<VNode[]>`


mightyiam:
[This CSS framework is asking me to put stuff directly in the body](http://bulma.io/documentation/layout/section/).

> "Use sections as direct children of body."

I can't seem to do that.

How about expanding DOMDriver to accept `Stream<VNode[]>`?

Steelfish:
If you are targetting body with the cycle dom driver, I believe you should be able to update it by passing a body vnode to the dom driver with the sections as its children.

TylorS:
```js
import { run } from '@cycle/run';
import { h, makeDOMDriver } from '@cycle/dom';

function main(sources) {
  return {
    DOM: xs.of(h('body', {}, [ ...children ]),
  }
}

run(main, { DOM: makeDOMDriver(document.body') })
```

mightyiam:
https://github.com/cyclejs/cyclejs/blob/c37c89d65a991f22f6459dc0fe8d578720ce8039/dom/src/VNodeWrapper.ts#L21
It seems that to pass that identity test the vnode must have a `.data.props.id`.

An easy hack I got working for me is to set it to an empty string.

So, that's great. It would be even less hacky if I didn't have to set that property.

staltz:
Weird, now that I look at VNodeWrapper, it sure looks like isVNodeAndRootElementIdentical is wrong/incomplete.

IssuehuntBot:
@issuehuntfest has funded $40.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/26316966/issues/540)

jvanbruegge:
The title is a bit misleading, this issue is basically a duplicate of https://github.com/cyclejs/cyclejs/issues/798

PR #792, opened 2018-03-23
Allow array of VDOM

bloodyKnuckles:
Allow array of VDOM objects to be passed to the DOMDriver sink. No need to wrap sibling elements in parent element. This is particularly useful when patching the HTML document HEAD.

I attempted to run tests but get this error: 
`test/node/mock-dom-source.ts (14,8): Cannot find module '../../lib/cjs/index'. (2307)`

https://github.com/cyclejs/cyclejs/blob/master/dom/test/node/mock-dom-source.ts#L14

I don't see that directory in the git repository.

### DIFF
--- a/dom/src/VNodeWrapper.ts
+++ b/dom/src/VNodeWrapper.ts
@@ -29,7 +29,7 @@ export class VNodeWrapper {
       return vnode;
     }
 
-    return this.wrap([vnode]);
+    return this.wrap(Array.isArray(vnode) ? vnode : [vnode]);
   }
 
   private wrapDocFrag(children: Array<VNode>) {


jvanbruegge:
The problem with arrays is isolation. We have to attach the isolation information to the vnode. So if we have no vnode, but an array that wont work

bloodyKnuckles:
Can the array of VDOMs automatically be wrapped with a single VDOM object only when isolation is needed?

jvanbruegge:
That would be implicit behavior, ie *sometimes* would happen *something* that is not expected, ie wrapping the array.
Explicit behavior is a lot better (aka wrapping it yourself)
Additionally parent components can be sure that they will always receive a single vnode and wont have to wrap everything in an `Array.isArray` if/else

mightyiam:
Perhaps the solution is a kind of merging of the parent component VNode with the child's single top VNode that would normally go within it? If they're in some form similar, do I remember correctly that this already could occur due to how snabbdom determines identity of VNodes? This might be how I put stuff directly in the body element.

bloodyKnuckles:
@mightyiam Are you saying DOMDriver would always take the provided container as the single top vnode? E.g.:

  function main () {
  }
  run(main, {
    DOM: makeDOMDriver('head')
  }


OR, would it be more explicit to allow me to provide:

`head([title('About Page'), script({attrs: {src: '/assets/client.js'}})])`

...and DOMDriver detect that and not duplicate the HEAD tag?

staltz:
Hi! If this is about the head section only, do you need to change the title, or actually everything in the head? I think for changing the title you could make a small driver that takes a stream of strings and changes the title node.

That said, I think we should consider how to fix the dom driver so that head inside head doesn't happen.

jvanbruegge:
Yeah, the element you send to the DOM driver should be the *same* as the one you attach it to. So
```ts
function main () {
  return {
    DOM: xs.of(head([
      title('About Page'),
      script({attrs: {src: '/assets/client.js'}})
    ]))
  }
}
run(main, {
  DOM: makeDOMDriver('head')
})
```
should result in
```
<head>
  <title>page title</title>
  <script src="/assets/client.js"></script> </head>
```
Therefore I would mark this as bug

jvanbruegge:
There is already #540 about that, so i will close this PR and as said check if is still the case in the new DOM driver

bloodyKnuckles:
The HEAD tag is still doubled with @cycle/dom v20.2.0. Here's an example:

https://bloodyknuckles.neocities.org/cyclehead/index.html
https://bloodyknuckles.neocities.org/cyclehead/bundlehead.js

jvanbruegge:
Yeah, I meant the upcoming v21.

bloodyKnuckles:
I don't know how to access v21 but in the currently available version if I change this line:

https://github.com/cyclejs/cyclejs/blob/master/dom/src/VNodeWrapper.ts#L16

From:

`const {tagName: selTagName, id: selId} = selectorParser(vnode);`

To:

`const {tagName: selTagName, id: selId = ''} = selectorParser(vnode);`

The head tag is no longer duplicated. 

The change is to give `selId` a default value of empty string to be comparable to the default value of the `rootElement`.

PR #794, opened 2018-03-26
Do Not Duplicate Root Elements w/o ID

bloodyKnuckles:
Do not duplicate the root element when not using ID property. For example, without this PR, when simply targeting the HEAD tag it gets duplicated because there is no associated ID. This PR simply gives the `selId` variable a default value of empty string, which correlate it with the default value of empty string given to the ID property of the `rootElement` object by `document.querySelector`.

Discussions here:
https://github.com/cyclejs/cyclejs/pull/792
https://github.com/cyclejs/cyclejs/issues/540

---------

- [ ] Has valid commit message
- [ ] Has tests

### DIFF
--- a/dom/src/VNodeWrapper.ts
+++ b/dom/src/VNodeWrapper.ts
@@ -13,7 +13,7 @@ export class VNodeWrapper {
     if (vnode === null) {
       return this.wrap([]);
     }
-    const {tagName: selTagName, id: selId} = selectorParser(vnode);
+    const {tagName: selTagName, id: selId = ''} = selectorParser(vnode);
     const vNodeClassName = classNameFromVNode(vnode);
     const vNodeData = vnode.data || {};
     const vNodeDataProps = vNodeData.props || {};


staltz:
Interesting that the fix is so simple! It would be nice to have a test for this fix, though. Anyone can do it, either you or me or Jan, but it's important to have. Otherwise it's easy to overlook the necessity of this small detail in case someone refactors this code in the future.

staltz:
I see that your patch branch is `patch-2` which most likely indicates you did the changes in GitHub website. That's not a problem, but to get everything right (tests and commit message), it's best if developing this locally then pushing. Do you mind if we close the PR and do it ourselves? If you want to have this as your own contribution with your name in the commits, then you could do the code changes locally on your computer.

bloodyKnuckles:
Thanks for the guidance! I'm glad to learn these tools. I forked, cloned, branched, fixed, tested, pushed, and pulled:

"Fix duplicate root element not having ID" bug PR: https://github.com/cyclejs/cyclejs/pull/796

:)

PR #795, opened 2018-03-26
Add duplicate root element fix test

bloodyKnuckles:
Adding a test to verify that root element is not duplicated when it's ID property is not set.

Test for PR https://github.com/cyclejs/cyclejs/pull/794

I'm not able to run tests in my CycleJS download. I get this error:
```
TSError: ⨯ Unable to compile TypeScript
test/node/mock-dom-source.ts (14,8): Cannot find module '../../lib/cjs/index'.
```

### DIFF
--- a/dom/test/browser/src/dom-driver.ts
+++ b/dom/test/browser/src/dom-driver.ts
@@ -231,4 +231,26 @@ describe('DOM Driver', function() {
       }, 50);
     }, 100);
   });
+
+  it('should not duplicate root element when not having ID property set', function(done) {
+    function app() {
+      return {
+        DOM: xs.of(div('dummy text')),
+      };
+    }
+
+    const {sinks, sources, run} = setup(app, {
+      DOM: makeDOMDriver(createRenderTarget()),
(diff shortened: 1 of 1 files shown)


jvanbruegge:
You have to run `make setup` first, so all cycle packages get compiled first
I will close this PR, could you add the test onto #794 directly as second commit?

bloodyKnuckles:
@jvanbruegge  Thanks for that! Now I can run tests. :)

I wrote a passing test and plan to send it as a new PR combined with the original fix.

I ran into a problem—the original fix (`setId = ''`), ended up breaking a couple other tests:
```
DOMSource.events() should catch bubbling events in a DocumentFragment
DOMSource.events() should catch non-bubbling events in a DocumentFragment with useCapture
```
These tests worked before because the root tag was duplicated since there was no ID assigned to either the root element nor the parent VDOM object. After fixing the "duplicate bug" the root tag was no longer duplicated causing the tests to fail. I added an ID to the root VDOM objects and the tests now pass.

PR #823, opened 2018-06-27
New run API: setupReusable

staltz:

In hacking together a way of using react-native-navigation with cyclejs, I had to build this:

```typescript
function setupReusable(drivers: Drivers): Engine {
  const sinkProxies = makeSinkProxies(drivers);
  const rawSources = callDrivers(drivers, sinkProxies);
  const sources = adaptSources(rawSources);
  const start = (sinks: Sinks) => replicateMany(sinks, sinkProxies);
  return {sources, start};
}
```

It's basically for reusing drivers without re-instantiating drivers, for multiple main() functions:

```javascript
const engine = setupReusable(drivers);
const sources = {
  ...engine.sources,
  screen: screensSource,
  navigation: navSource,
};
const sinks = main(sources);
engine.start(sinks);
```

It's like a partially-applied variant of `setup`:

```
setup(main, drivers): {sources, sinks, run: () => Dispose}
setupReusable(drivers): {sources, run: (sinks) => Dispose}
```

And I prefer a longer name like that so to signal to people that it's not as common as `run(main, drivers): Dispose`.

------

How to review this PR: the [`feat` commit](https://github.com/cyclejs/cyclejs/pull/823/commits/d58947ad908bf5fb17aabb07e91629e6ec36bd73) is the important one, the `refactor` commit is just about moving code around.


run/src/index.ts
```diff
@@ -229,6 +231,63 @@ export function setup<So extends Sources, Si extends FantasySinks<Si>>(
...
+ * contains the generated sources (from those drivers) and a `run` function
+ * (which in turn expects sinks as argument). This `run` function can be called
+ * multiple times with different arguments, and it will reuse the drivers that
+ * were passed to `setupReusable`.
+ *
+ * **Example:**
+ * ```js
+ * import {setupReusable} from '@cycle/run';
+ * const {sources, run} = setupReusable(drivers);
+ * const sinks = main({...sources, foo: fooDriver});
```
jvanbruegge: I would assume you can't pass in a driver directly as sources, do you?

staltz: Good point!

staltz:
Fixed!

Issue #826, opened 2018-06-30
HTTP isolation expects 'map' by default, Rxjs6 does not provide it by default


ntilwalli:
in Rxjs6 all operators are applied through pipe. When not importing ‘rxjs-compat’ the map operator is not available on the base observable. Isolation currently assumes map exists on the streams sent as sinks. This means importing ‘rxjs-compat’ is effectively required to ensure the map operator exists on the sink observables... is the best approach to fork cycle/isolate for myself and apply via pipe? Is there something more elegant?

jvanbruegge:
I think the cleanest approach would be to `adapt` in `@cycle/isolate`

staltz:
Are you sure this is a problem in cycle/isolate? I can't even find the map operator in the source code of isolate

ntilwalli:
Ah, I saw that isolation was the issue and jumped to the broadest conclusion, my bad, it's in HTTP: https://github.com/cyclejs/cyclejs/blob/master/http/src/isolate.ts#L38

jvanbruegge:
Yeah, I would assume the driver is written with xstream, so `@cycle/isolate` should `adapt` before calling into a driver

ntilwalli:
Are you suggesting that a second function, lateral to `adapt` be registered?   Like a function indicating how to apply a 'map' operation so `isolate` can apply that registered function instead of calling 'map' directly?  

In this case the HTTP driver does `request$.map(someFn)` but in the case of Rxjs6 it should import the `map` operator from `rxjs/operators` and call `request$.pipe(map(someFn).  So I guess in the registeration approach it would do something like `applyMapFromAdapt(request$, someFn)`?  Do I understand correctly?

staltz:
To be more accurate, `@cycle/isolate` should `adapt` *after* calling into a driver's isolateSink.

Reminder: `adapt` is xstream=>others. `xs.fromObservable` is others=>xstream.

HTTP driver is written internally with xstream. So when `@cycle/isolate` call's HTTP's isolateSink, it must pass an xs stream as the `innerSink` to it, *or* HTTP driver itself converts using xs.fromObservable, then uses map, and returns an xs stream. `@cycle/isolate` gets that xs stream, calls `adapt` on it and gets the `outerSink` usable by your application.

jvanbruegge:
[...]

PR #827, opened 2018-07-04
fix(isolate): adapt sinks so xstream semantics can be applied

ntilwalli:
Rxjs 6 uses a piping semantic which removes the existence of the map prototype from the base
Observable class, this change ensures that each sink stream is adapted to xstream before
isolation
operations are applied.

BREAKING CHANGE:
This update changes the contract such that xstream and @cycle/run are now dependencies.

ISSUES CLOSED: #826


staltz:
Thanks @ntilwalli! There are some minor things to tweak, but I'll do that on my own, I won't bother you with them :)

PR #836, opened 2018-08-17
refactor(META): delete unused scripts

jvanbruegge:
Not useful any more due to `pnpm recursive run`


jvanbruegge:
I am not sure about the symlink script, is it used in the docs @staltz ?

Issue #713, opened 2017-10-09
DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.


whitecolor:
Here is a bug I've met with simple list and items selections:

**Code to reproduce the issue:**

```js
import xs from 'xstream'
import { run } from '@cycle/run'
import { makeDOMDriver, div } from '@cycle/dom'

console.clear()

const Main = ({DOM}) => {
   const selectedIndex$ = DOM.select(`[index]`).events('click')
    .map(ev => parseInt((ev.currentTarget).getAttribute('index')))
    .startWith(0)

   const selectedDOM$ = xs.of(div(['Selected']))
  
  return {
    DOM: xs.combine(selectedIndex$, selectedDOM$)
      .map(([selectedIndex, selectedDOM]) => {
        const list = ['One', 'Two', 'Three']
        return div(
          list.map((item, index) => {
            const isSelected = index === selectedIndex
            return div({
              attrs: {index}
            }, [isSelected ? selectedDOM : item ])
          })
        )
      })
  }
}

run(Main, {
  DOM: makeDOMDriver('body')
})
```

**Expected behavior:**

When clicked the item (`One`, `Two`, or `Three` ) is replaced with `Selected` text.

**Actual behavior:**

After a few clicks:

> DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.

![image](https://user-images.githubusercontent.com/736697/31338304-1a82d778-ad18-11e7-8b41-cbfd9fe5ea99.png)

**Versions of packages used:**

![image](https://user-images.githubusercontent.com/736697/31338315-22142226-ad18-11e7-87ba-c72c9a420617.png)

Here is webpackbin:
https://www.webpackbin.com/bins/-Kw01D616yUWznH1mbTA

staltz:
Thanks for the detailed report

brucou:
I had that bug twice too. I fixed it by removing the array. i.e. run your code with replacing `div({              attrs: {index}            }, [isSelected ? selectedDOM : item ])`

by `div({              attrs: {index}            }, isSelected ? selectedDOM : item )`

Now the actual reason of the bug I don't know. I did not investigate further, snabbdom source code is pretty hard to read.

whitecolor:
@brucou right that works

This is supposedly snabbdom problem.

staltz:
[...]

PR #843, opened 2018-09-02
chore(dom): add test for #713

jvanbruegge:
Test is currently failing, will investigate

- [x] There exists an issue discussing the need for this PR #713 


jvanbruegge:
This is by design, I added a section in the README of snabbdom to make this clear

PR #844, opened 2018-09-20
Update to typescript 3.0, release @cycle/state

jvanbruegge:


staltz:
So if I'm not mistaken, this is what we should do?

- [ ] Release run@5 and isolate@4
- [ ] Run CI tests where packages depend on run@5 and isolate@4
- [ ] Release new versions for all other packages (except state)
- [ ] Merge PR, make release notes public on master branch
- [ ] Update cycle/state's API (at least onionify=>withState)
- [ ] Release cycle/state

jvanbruegge:
Yes, in that order

staltz:
Updated the todo list a bit. Also check the "chore(META)" commits

staltz:
@jvanbruegge I fixed some small issues but now the DOM tests exploded with TS errors in CI

jvanbruegge:
The typescript issue is a rabbit hole. The main issue is that we return `MainDOMSource` from the driver and not ´DOMSource`, but changing this leads to a ton of other errors

jvanbruegge:
`MockedDOMSource` is commented, because its methods return any, which means that the union type returns any

jvanbruegge:
The cycle/run changes are needed to make everything compile

run/src/types.ts
```diff
@@ -27,24 +27,24 @@ export type Driver<Si, So> = {
...
 
 export type Sinks<M extends Main> = {
-  [k in (string & keyof ReturnType<M>)]: ReturnType<M>[k] & FantasyObservable
+  [k in string & keyof ReturnType<M>]: ReturnType<M>[k]
 };
 
 export type MatchingMain<D extends Drivers, M extends Main> = Main & {
-  (so?: Partial<Sources<D>>): Sinks<M>;
+  (so: Sources<D>): Sinks<M>;
 };
```
staltz: This change is the one that worries me the most, why was it needed? What compile error happens in dom if we revert it?
It's important that we can support `main` functions that either don't use sources or use some of the sources, but this code change means that `main` will have to specify sources always.

run/src/types.ts
```diff
@@ -27,24 +27,24 @@ export type Driver<Si, So> = {
 export type DisposeFunction = () => void;
 
 export type Drivers = {
-  [name: string]: Driver<FantasyObservable | undefined | void | null, any>;
+  [name: string]: Driver<Stream<any> | undefined, any>;
```
staltz: The change from FantasyObservable to Stream<any> means that it's going from more general to more specialized. It may work for xstream, but we shouldn't require anything else than the "subscribe" method. Also curious what compile error happens if we revert this change. Remember that we have already released the new cycle/run with TypeScript support, so we're risking issuing yet another breaking change with this code change.

isolate/package.json
```diff
@@ -31,7 +31,7 @@
   "typings": "lib/cjs/index.d.ts",
   "types": "lib/cjs/index.d.ts",
   "dependencies": {
-    "@cycle/run": "^5.0.0",
+    "@cycle/run": "link:../run",
```
staltz: Is this temporary?

isolate/package.json
```diff
@@ -46,7 +46,7 @@
     "symbol-observable": "^1.2.0",
     "ts-node": "^7.0.0",
     "tslint": "^5.11.0",
-    "typescript": "3.0.x"
+    "typescript": "^3.1.3"
```
staltz: Can we first update to TS 3.0 then later to 3.1? I think 3.1 introduced a few breaking changes.

dom/test/browser/events.ts
```diff
@@ -885,15 +886,15 @@ describe('DOMSource.events()', function() {
...
+    assert.strictEqual((event$ as any)._isCycleSource, 'DOM');
     done();
   });
 
   it('should allow restarting of event streams from isolated components', function(done) {
     const outSubject = xs.create<any>();
     const switchSubject = xs.create<any>();
 
-    function component(_sources: {DOM: DOMSource}) {
+    function component(_sources: {DOM: MainDOMSource}) {
```
staltz: Why is this needed? A change in tests usually indicates a breaking change and migration notes

staltz:
I guess I don't understand why the move from interface to union type. Conceptually it seems that the interface would be more adequate, because it just describes the "type class" with the methods that we need in the `main`, but as a union type we're instead doing a switch between different implementations that all have something in common.

run/src/types.ts
jvanbruegge: I have to try again (in case this was not needed in the end) but AFAICR it threw an error if you specified all sources
Maybe a union of both solves this

run/src/types.ts
jvanbruegge: But isnt this more correct? A driver gets and has to return xstream after all

jvanbruegge: But isnt this more correct? A driver gets and has to return xstream after all

isolate/package.json
jvanbruegge: yes

isolate/package.json
jvanbruegge: I read through it and i didnt spot any

run/src/types.ts
staltz: Yeah I think you're right: https://cycle.js.org/releases.html#cycle-unified-for-library-authors-driver-authors

dom/test/browser/events.ts
jvanbruegge: This was by mistake

isolate/package.json
jvanbruegge: Had a look, the only breaking changes are
- Some vendor-specific types are removed from lib.d.ts (ie element.msMatchesSelector, not relevant for the user)
- Some [typeof](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#narrowing-functions-now-intersects--object-and-unconstrained-generic-type-parameters) changes that also only affect us internally because they are backwards compatible with 3.0

jvanbruegge:
Fixed everything! Was able to go back to the interface and I added a test to `@cycle/run` that checks for main with partial sources and no sources.

jvanbruegge:
I also released canary versions for `@cycle/run`, `@cycle/rxjs-run` and `@cycle/isolate` so CI should not fail now

jvanbruegge:
This is done now, ready for review and publish

jvanbruegge:
One of the tests is flaky in android, the one about cross driver race condtitons

staltz:
Reviewing...

staltz:
Note to self: 

- [ ] add tests "should allow to not use all sources in main" to most-run and rxjs-run
- [ ] document migration for rxjs-typings and most-typings (dom and http)

staltz:
Note that the `chore` commits you just made should have been `fix` commits with breaking changes (that will show in the changelog), specially regarding rxjs-typings etc. I will take care of this, by doing a big rebase -i with the rest of the `fix` commits from the past.

jvanbruegge:
Yeah, I thought the new commits should be squashed together with the old ones

jvanbruegge:
I would disable the test in Android for now, I tried debugging this issue, but when run live on android the error does not occur.

staltz:
I'm hopeful that either splitting the test in two, or increasing the timeouts, would solve it

mightyiam:
Thank you for all of the work you're putting into this project! Thank you!

Issue #847, opened 2018-10-06
Typings (e.g. run and dom but probably others) don't work as expected


texastoland:
I was reviewing the [egghead course](https://github.com/eggheadio-projects/cyclejs-fundamentals) and tried to add typings since the files are `*.ts`. Starting with the latest dependencies, `"strict": true` in `tsconfig.json`, and the [counter example](https://github.com/cyclejs/cyclejs/blob/master/examples/basic/counter/src/main.js):

```js
function main(sources) {
  // see link for full implementation
  return {
    DOM: vdom$
  };
}

run(main, {
  DOM: makeDOMDriver('#app')
});
```

My intuition was it should just work with a single annotation on `main`:

```ts
function main(sources: {DOM: DOMSource}) {
```

The first problem is `run`'s types don't match because `makeDOMDriver` returns its concrete type (`MainDOMDriver`) instead of interface. I ran into this with with XStream's `MemoryStream` as well. Not ideal but easy to fix.

```ts
function main(sources: {DOM: MainDOMSource}) {
```

The next problem is `run` expects `main` to handle `undefined`. The signature is easy to fix.

```ts
function main(sources?: { DOM?: MainDOMSource }) {
```

How `main` should recover from missing sources is less clear. Not ideal but fixable with non-null assertions e.g. `sources!.DOM!.select`.

The final problem is `main` returns `Stream<VNode>` but `run` expects the driver's stream to be either `FantasyObservable | undefined | void | null`. `null` seems like a logic error. `void` and `undefined` are synonyms already inferred by `Driver`. `Stream` is not a `FantasyObservable` because the latter expects `subscribe` to handle `undefined`. Not easy to fix.

The typings look impressive. They just appear underused. Converting the existing examples to TypeScript with as minimal annotations as possible would reveal these usability issues.

fasiha:
Possibly related: https://github.com/staltz/cycle-onionify/issues/69 I found that the `bmi-typescript` example fails to build with TypeScript 2.5 or newer.

Sinewyk:
Code
```ts
import { run } from '@cycle/run';
import { makeHTTPDriver } from '@cycle/http';

function main() {
    return {};
}

const drivers = {
  HTTP: makeHTTPDriver(),
};

run(main, drivers);

```

Error
```
[...]

PR #852, opened 2018-10-18
fix(run): change signature of a driver function

jvanbruegge:
ISSUES FIXED: #847 

A simple example did not work without this change, and this is the more correct signature


Sinewyk:
Anyway we can help ?

jvanbruegge:
@staltz now properly rebased

dom/src/utils.ts
```diff
@@ -73,11 +73,11 @@ export function isEqualNamespace(
 }
 
 export function makeInsert(
-  map: Map<string, Map<Element, any>>,
+  map: Map<string, Map<Element, any>>
 ): (type: string, elm: Element, value: any) => void {
   return (type, elm, value) => {
     if (map.has(type)) {
-      const innerMap = map.get(type) as Map<Element, any>;
+      const innerMap = map.get(type)!;
```
staltz: What's this? :eyes:

dom/test/node/mock-dom-source.ts
```diff
@@ -228,7 +230,7 @@ describe('isolation on MockedDOMSource', function() {
     }
 
     const {sinks, sources, run} = setup(app, {
-      DOM: () =>
+      DOM: (_: Stream<VNode>) =>
```
staltz: Was this added to please TypeScript? I think we should still support "ignored main() sinks" so the inference is not strict about all main() sinks needing to have a respective sink-receiving driver. A change in tests indicates a possible breaking change too

dom/src/ElementFinder.ts
```diff
@@ -10,7 +10,7 @@ function toElArray(input: any): Array<Element> {
 export class ElementFinder {
   constructor(
     public namespace: Array<Scope>,
-    public isolateModule: IsolateModule,
+    public isolateModule: IsolateModule
```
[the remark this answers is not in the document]
jvanbruegge: We had to change the trailing commas to ES5 compatible a [while back](https://github.com/cyclejs/cyclejs/commit/40fcca120b7996316a638856db7f0eb7a51444ba#diff-b9cfc7f2cdf78a7f4b91a753d10865a2), but the formatting glob was a bit broken, so a few files were spared the reformat

dom/src/utils.ts
jvanbruegge: a non-null assertion. `get` returns `x | undefined`, this is to force the result to be just `x`. The previous did the same, but for some reason tslint complains that it does not change the type of the expression.

dom/test/node/mock-dom-source.ts
jvanbruegge: Are you sure? I would want a type error if a driver would ignore the sinks I return from main. This usually means i did something wrong

dom/src/utils.ts
staltz: wow, this is really useful, is it how new in TypeScript?

dom/test/node/mock-dom-source.ts
staltz: Yes I'm quite sure this should be an allowed use case. Simplest scenario is taking a main() and testing only one channel of it. Or running only one of its channels. It improves the reusability of main() because it doesn't couple the main with any particular subset of sinks dictated by the drivers object.

dom/src/utils.ts
jvanbruegge: since TypeScript 2.0 apparently

dom/test/node/mock-dom-source.ts
jvanbruegge: Isnt it the other way round? With this type the drivers object is dictated by the main function. Main is allowed to ignore drivers, but drivers are not allowed to ignore main

staltz: Yes, drivers are allowed to ignore *some channels* from main.

dom/src/DOMSource.ts
```diff
@@ -20,4 +25,4 @@ export interface DOMSource {
     bubbles?: boolean
   ): Stream<HTMLElementEventMap[K]>;
   events(eventType: string, options?: EventsFnOptions): Stream<Event>;
-}
+}*/
```
staltz: Could you remove this comment?

dom/src/DOMSource.ts
```diff
@@ -8,7 +11,9 @@ export interface EventsFnOptions {
   preventDefault?: PreventDefaultOpt;
 }
 
-export interface DOMSource {
+export type DOMSource = MainDOMSource | DocumentDOMSource | BodyDOMSource; // | MockedDOMSource;
```
staltz: Any reason why MockedDOMSource is not included? (it's okay if it's commented, as long as we write down the reason why it's excluded, in a comment)

http/src/http-driver.ts
```diff
@@ -117,8 +118,8 @@ export function createResponse$(reqInput: RequestInput): Stream<Response> {
       }
     },
     stop: function stopResponseStream() {
-      if (this.request && this.request.abort) {
-        this.request.abort();
+      if (request && request.abort) {
+        request.abort();
```
staltz: I just noticed that in both the `this` implementation as with this new closure `request` implementation, we should reset the request back to undefined or null in this `stop()` method, so that the next execution of this stream doesn't accidentally call "abort" on the request of the previous execution. It's a rare corner case that may not happen, but since we're at it, might be good to add `request = null` just in case.

staltz:
You could rebase "fixup" the commit `chore(dom): revert changes to mockDOMSource tests` with the previous `fix(dom): ...` commit

staltz:
All good!

PR #856, opened 2018-10-19
fix(dom): add better typings for the hyperscript helpers

jvanbruegge:

### DIFF
--- a/dom/src/hyperscript-helpers.ts
+++ b/dom/src/hyperscript-helpers.ts
@@ -1,6 +1,6 @@
 // tslint:disable:max-file-line-count
 import {h} from 'snabbdom/h';
-import {VNode} from 'snabbdom/vnode';
+import {VNode, VNodeData} from 'snabbdom/vnode';
 
 function isValidString(param: any): boolean {
   return typeof param === 'string' && param.length > 0;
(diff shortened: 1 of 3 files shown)


jvanbruegge:
These bad typings are the reason why [the zombie transposition](https://github.com/cyclejs/cyclejs/commit/aff1c9dc46603a2e0935dda4b5aa7aaa943fde7f) could survive for so long in the tests

staltz:
Looks good. I have just one question about this: is it a breaking change? Technically because people had `any` types before, with this change, their TypeScript compiler might light up with many errors.

jvanbruegge:
It should not be as all valid uses of the hyperscript helpers are captured in the types. So if this now throws an error where it did not before, then because of a misuse of the API (like transposition).

jvanbruegge:
one nice thing to do would be to merge https://github.com/snabbdom/snabbdom/pull/325 first, to remove the `any` in the typings

jvanbruegge:
@staltz the snabbdom PR may take ages to be released, let's merge and release this now with any, we can still improve the typings in the future

staltz:
Okay, how about CI?

jvanbruegge:
Should be fixed now

jvanbruegge:
yes, CI is green

staltz:
Can you merge? I'm on mobile

Issue #848, opened 2018-10-09
@cycle/time operators does not always complete properly


fuunnx:
Hi !

I had a memory leak in my app and I found that the root cause were `Time` operators not completing properly (which remained active and triggered heavy computations multiple times). 

Adding this statement here https://github.com/cyclejs/cyclejs/blob/master/time/src/debounce.ts#L59 fixed it :
```js
stop() {
  stream.shamefullySendComplete()
},
```

I'll add a reproductible example later and would be glad to make a pull request to fix it of course

fuunnx:
Hi, as promised here is a minimal reproducible example :

[![Edit yj4x45l4oz](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/yj4x45l4oz)

**Expected behavior** : No more 'before' logs after 1200ms when using a `Time` operator

**Observed behavior** : They are still there. They disappear as expected when no more `Time` operator is in the chain.

Widdershin:
@fuunnx thanks for your report and reproducer. I'm unlikely to have time to sort this out soon so if you do want to make a pull request that would be fantastic.

fuunnx:
I'll do, I've already started :)

PR #858, opened 2018-10-29
fix(time): unsubscribes Time operators on complete

fuunnx:
ISSUES CLOSED: #848

### DIFF
--- a/time/src/debounce.ts
+++ b/time/src/debounce.ts
@@ -48,10 +48,11 @@ function makeDebounce(createOperator: () => OperatorArgs<any>) {
     return function debounceOperator<T>(inputStream: Stream<T>): Stream<T> {
       const state = {scheduledEntry: null};
       const stream = xs.fromObservable(inputStream);
+      let debounceListener: any = null;
 
       const debouncedStream = xs.create<T>({
         start(listener: Listener<T>) {
-          const debounceListener = makeDebounceListener<T>(
+          debounceListener = makeDebounceListener<T>(
             schedule,
             currentTime,
             debounceInterval,
@@ -63,7 +64,9 @@ function makeDebounce(createOperator: () => OperatorArgs<any>) {
         },
 
         stop() {
-          stream.shamefullySendComplete();
+          if (debounceListener) {
+            stream.removeListener(debounceListener);
+          }
         },
       });
 
--- a/time/src/debounce.ts
+++ b/time/src/debounce.ts
@@ -26,10 +26,16 @@ function makeDebounceListener<T>(
     },
 
     error(e: any) {
+      if (state.scheduledEntry) {
+        state.scheduledEntry.cancelled = true;
+      }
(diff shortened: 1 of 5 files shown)


fuunnx:
Also while reviewing the code I found this TODO https://github.com/cyclejs/cyclejs/blob/master/time/src/time-driver.ts#L113 which seems easy to fix and in the scope of this pull request

time/test/time.ts
```diff
@@ -50,6 +50,41 @@ function map(stream: any, f: any) {
   return stream.map(f);
 }
 
+function testUnsubscription(Time: TimeSource, operator: any, done: Mocha.Done) {
+  const PERIOD = 20;
+  const taps: Array<number> = []; // tslint:disable-line
```
jvanbruegge: why is the tslint disabled?

fuunnx: oh yeah it's not necessary here anymore
I had declared 
`let taps: Array<number> = [];`  but then tslint complained : 
`Identifier 'taps' is never reassigned; use 'const' instead of 'let'.`

fuunnx:
Yep, the tests fail without my changes =)

time/src/debounce.ts
```diff
@@ -52,11 +59,12 @@ function makeDebounce(createOperator: () => OperatorArgs<any>) {
...
           );
 
-          xs.fromObservable(stream).addListener(debounceListener);
+          stream.addListener(debounceListener);
         },
 
-        // TODO - maybe cancel the scheduled event?
-        stop() {},
+        stop() {
+          stream.shamefullySendComplete();
```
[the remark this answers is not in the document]
fuunnx: it works, indeed
I update the PR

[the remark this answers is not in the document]
fuunnx: Updated :)

time/test/time.ts
fuunnx: Updated :)

jvanbruegge:
LGTM, @Widdershin do you want to have another look?

Widdershin:
Looks good to me 😊 thanks a ton @fuunnx, much appreciate you taking the lead on this.

Only consideration is that I don't have time to test this, so if someone could double check that the examples still work and if you could confirm that it resolves the memory leak in your app @fuunnx, I would much appreciate it. That aside, merge when ready 🐈:shipit:

fuunnx:
1) I confirm that the memory leak is no longer reproducible in my app
2) `stream.shamefullySendComplete();` was indeed really dangerous and caused weird behaviours in my app 
3) What are the examples to test ?

Widdershin:
There's just one example in `time/example`, I checked and it works fine 😄 

Thanks for the great job you did on this PR!

Widdershin:
@jvanbruegge are you happy to merge and publish?

jvanbruegge:
Thanks again for the PR!

Issue #860, opened 2018-11-27
Wrong typings for RxJS and Most.js versions of makeDOMDriver


kosmogradsky:
@cycle/dom v22.0.0 introduced separate makeDOMDriver() functions for RxJS and Most.js, but Typescript throws an error when I try to use them. Probably because the new separate functions describe no arguments in their types:
https://github.com/cyclejs/cyclejs/commit/8174c25#diff-6ec374a820c34eab57b27ca30d8e6ad8R22
https://github.com/cyclejs/cyclejs/commit/8174c25#diff-01aea6cd37e196dd39a5549985abd670R18

**Code to reproduce the issue:**
```
import { h1, VNode } from '@cycle/dom';
import { makeDOMDriver, DOMSource } from '@cycle/dom/lib/cjs/rxjs';
import { run } from '@cycle/rxjs-run';
import { of, Observable } from 'rxjs';

import './styles.css';

type mainFunction = (sources: {
  DOM: DOMSource
}) => {
  DOM: Observable<VNode>
};


const main: mainFunction = () => {
  const vdom$ = of(h1('404'));

  return {
    DOM: vdom$
  };
};

const drivers = {
  DOM: makeDOMDriver('#root')
};

run(main, drivers);
```

**Expected behavior:**
Code above compiling without any errors.

**Actual behavior:**
Typescript throws an error `Expected 0 arguments, but got 1.` when I pass `'#root'` to `mainDOMDriver()`.

**Versions of packages used:**
```
"@cycle/dom": "^22.0.0",
"@cycle/history": "^7.0.0",
"@cycle/rxjs-run": "^10.1.0",
"typescript": "^3.1.6",
"rxjs": "^6.3.3",
```

kosmogradsky:
This can be worked around using `makeDOMDriver()` from `@cycle/dom` and type assertion.

```
import { DOMSource } from '@cycle/dom/lib/cjs/rxjs';
import { makeDOMDriver, VNode } from '@cycle/dom';
import { Stream } from 'xstream';

const drivers = {
  DOM: makeDOMDriver('#root') as unknown as Driver<Stream<VNode>, DOMSource>
};
```

jvanbruegge:
Yeah, sorry, this is an incredibly stupid bug. Will fix that tomorrow
One more reason to figure out how to do #802

kosmogradsky:
I've also encountered some problems with `DOMSource` interface for Most.js and RxJS:
1. `element()` method is missing
[...]

PR #862, opened 2018-12-04
fix(dom): fix rxjs and most typings of makeDOMDriver

jvanbruegge:
ISSUES CLOSED: #860

### DIFF
--- a/dom/src/most.ts
+++ b/dom/src/most.ts
@@ -1,22 +1,22 @@
 import {EventsFnOptions} from './DOMSource';
-import {makeDOMDriver as make, VNode} from './index';
+import {makeDOMDriver as make, VNode, DOMDriverOptions} from './index';
 import {Stream} from 'most';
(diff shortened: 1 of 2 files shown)


dom/src/most.ts
```diff
@@ -1,22 +1,22 @@
...
-    eventType: string,
-    options?: EventsFnOptions,
-    bubbles?: boolean
-  ): Stream<Event>;
+  events(eventType: string, options?: EventsFnOptions): Stream<Event>;
 }
 
-export const makeDOMDriver: () => Driver<xs<VNode>, DOMSource> = make as any;
+export const makeDOMDriver: (
+  sel: string,
```
staltz: I think this argument is `container: string | Element`

jvanbruegge: Yes, updated that

jvanbruegge:
ok

PR #867, opened 2018-12-14
Add mockHTTPDriver helper

Sinewyk:
Here is the rebased and up to date helper to test the HTTP driver.

Does it also need to be tested inside the browser ?

- [x] There exists an issue discussing the need for this PR #567 


jvanbruegge:
No, mockHTTP source is just like mockDOM source intended for node testing

Sinewyk:
Kind of missed the point then. Not sure what the difference is, will check later.

Sinewyk:
Talked about it on gitter, may be correct ?

Tagging @Widdershin because you may be interested.

Sinewyk:
Closing because callbag rewrite, and all these issues _should_ disappear.

Issue #294, opened 2016-04-23
HTTP: attach for file uploads does not work in the browser


dmitriid:
Current implementation of `attach` does not work in the browser with File objects retrieved from `<input type='file' />`

The reason is that due to security reasons `path` and `filename` properties are not available to file objects:

```
> fileInput
<input type=​"file" multiple class=​"file">​

> fileInput.files
FileList {0: File, length: 1}

>  fileInput.files[0]
File {
  name: "1324373_original.jpg", 
  lastModified: 1460017515000, 
  lastModifiedDate: Thu Apr 07 2016 10:25:15 GMT+0200 (CEST), 
  webkitRelativePath: "", 
  size: 184841,
  _proto_: Blob
}
```

So, if we try to do this:

```js
return {
  HTTP: Rx.Observable.just({
    url: '/upload',
    attach: [fileInput.files[0]],
    headers: {'Content-Type': 'multipart/form-data'}
  })
}
```

This will fail due to:

```js
  if (attach !== null) {
    for (let i = attach.length - 1; i >= 0; i--) {
      const a = attach[i]
      request = request.attach(a.name, a.path, a.filename)
    }
  }
```

`a.path` and `a.filename` are `undefined` and will cause a crash in Superagent code.

A possible solution: allow `formData` in options/`attach`. Using code from  [here](https://blog.gaya.ninja/articles/uploading-files-superagent-in-browser/)

```js
// vanilla JS

var files = document.getElementById('file_to_upload').files;
var formData = new FormData();

for (var key in files) {
    // is the item a File?
    if (files.hasOwnProperty(key) && files[key] instanceof File) {
        formData.append(key, files[key]);
    }
}

superagent.post('/some-url/')
  .send(formData)
  .end(function(err, response) {
    //upload done!
  });
```

Possible usage with HTTP:

```js
return {
   HTTP: Rx.Observable.just({
      url: '/upload',
      formData: formData
   })
}
```

dmitriid:
Or a five-line file upload driver :)

```js
const makeFileDriver = (opts) => {
  const subj = new Rx.Subject()
  const xhr  = new XMLHttpRequest()
  xhr.open('POST', opts.location, true)

  xhr.onload = () => {
    if (xhr.status === 200) {
      subj.onNext({type: 'upload', success: true})
    } else {
      subj.onError({type: 'upload', success: false})
    }
  }

  return (files$) => {
    files$.map((files) => {
      const formData = new FormData()
[...]

PR #868, opened 2018-12-22
Add file upload example and address issue #294 with solution.

JuniperChicago:


staltz:
Hi @JuniperChicago, thanks for the initiative. Is this a work in progress PR? There are a couple things to do before merging:

- Squash all commits into one and use `pnpm run commit`
- Remove commented code
- Reformat code using prettier, like in the rest of the codebase
- Remove CSS styles that make this example differ from other examples
- etc (I'll review more, also Jan might review)

PR #873, opened 2019-01-25
WIP use TypeScript `unknown` in isolate

staltz:
**Do not merge yet**

This PR just shows *what kind of breaking changes* would happen if we would introduce TypeScript's `unknown` instead of `any`. The purpose is to avoid sneaking `any` types that tend to virally propagate, and instead to use `unknown` which forces the developer to cast the type of some sinks returned from an isolated component. In other words, `isolate` would never give `any` types, it would force the developer to think about what type they expect.

Open for discussion before I do the following:

- [x] rebase/split commits into one commit per package
- [x] release new isolate, breaking change, with migration guide
- [x] update each package to use the new isolate

### DIFF
--- a/isolate/test/index.ts
+++ b/isolate/test/index.ts
@@ -440,7 +440,7 @@ describe('isolate', function() {
       }
 
       function MyDataflowComponent(
-        sources: {other: any},
+        sources: {other: unknown},
         foo: string,
         bar: string
       ) {
@@ -456,8 +456,8 @@ describe('isolate', function() {
       );
 
       assert.strictEqual(typeof scopedSinks, `object`);
-      scopedSinks.other.subscribe((x: Array<string>) => {
-        assert.strictEqual(x.join(), `foo,bar`);
+      scopedSinks.other.subscribe(strings => {
+        assert.strictEqual(strings.join(), `foo,bar`);
         done();
       });
     });
@@ -492,11 +492,11 @@ describe('isolate', function() {
         {other: new MyTestSource()},
         `foo`,
         `bar`
-      );
+      ) as {other: Observable<Array<string>>};
 
       assert.strictEqual(typeof scopedSinks, `object`);
-      scopedSinks.other.subscribe((x: Array<string>) => {
-        assert.strictEqual(x.join(), `foo,bar`);
+      scopedSinks.other.subscribe(strings => {
+        assert.strictEqual(strings.join(), `foo,bar`);
         done();
       });
     });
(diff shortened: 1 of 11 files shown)


isolate/test/index.ts
```diff
@@ -440,7 +440,7 @@ describe('isolate', function() {
       }
 
       function MyDataflowComponent(
-        sources: {other: any},
+        sources: {other: unknown},
```
staltz: Even before this PR, this line should have been `unknown`.

But now, with this PR, this test does not pass (type check) with `any` (because `scopedSinks.other` would have been inferred as unknown, which would be a compile error a few lines below), so I had to change it to `unknown`.

isolate/test/index.ts
```diff
@@ -492,11 +492,11 @@ describe('isolate', function() {
         {other: new MyTestSource()},
         `foo`,
         `bar`
-      );
+      ) as {other: Observable<Array<string>>};
```
staltz: If Isolate were aware of Most.js and RxJS types of streams, we could support this test to type check without any changes. So we have to manually cast to RxJS's stream type here, otherwise it would be `unknown`.

isolate/test/index.ts
```diff
@@ -531,11 +531,11 @@ describe('isolate', function() {
         {other: new MyTestSource()},
         `foo`,
         `bar`
-      );
+      ) as {other: Observable<Array<number>>};
```
staltz: Same case as above.

isolate/test/index.ts
```diff
@@ -624,15 +624,15 @@ describe('isolate', function() {
         };
       }
 
-      function MyDataflowComponent(sources: {other: any}) {
+      function MyDataflowComponent(sources: {other: unknown}) {
```
staltz: Same case as [this comment](https://github.com/cyclejs/cyclejs/pull/873/files#r251171675)

state/test/withState.ts
```diff
@@ -142,7 +142,7 @@ describe('withState', function() {
     function Parent(sources: ParentSources): ParentSinks {
       const childSinks = isolate(Child, {state: 'child'})(sources);
       return {
-        state: childSinks.state,
+        state: childSinks.state as Stream<Reducer<Parent>>,
```
staltz: This case is a true example of an inference that we don't yet know how to support. Previously Isolate was inferring these cases as either `Stream<Reducer<any>>` or `Stream<Reducer<{}>>`, with this PR, the type is `Stream<unknown>` so it breaks on purpose unless the developer does a type cast, like we're doing here.

Issue #845, opened 2018-09-24
Updating most.js in cyclejs/most-run to @most/core


kanitsharma:
@most/core is pretty stable right now and even encouraged by the people at most.js to be used in new projects.

![image](https://user-images.githubusercontent.com/23506120/45982625-11f96b80-c077-11e8-8c38-1f6389a789c8.png)

plus it provides curried API and a tree shakable package which would be awesome to have :smile:

jleider:
I second this, are there any possible updates?

IssuehuntBot:
@staltz has funded $40.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/26316966/issues/845)

jvanbruegge:
I had a look at this, but `@most/core` feels still very alpha. There are barely any operators, for example the important `from` operator is completely missing

janat08:
Using @most/create since it's backwards compatible I changed adapt and added test to see if @most/core actually runs. It didn't run, while the old tests run with mostjs V1.
```
  1) run()
       supports mostjs core, the foundation for version 2:
     TypeError: this.ins.subscribe is not a function
      at FromObservable._start (node_modules/xstream/src/index.ts:134:26)
```
``` .ts
  it('mostjs core can be used in cyclejs', function (done) {
    const sandbox = sinon.createSandbox();
    const spy = sandbox.spy();
    const mainSpy = sandbox.spy()
    function app(sources: any) {
      const other$ = sources.other
      const spy$ = tap(mainSpy, other$)
      const sink$ = startWith('a', spy$)
      return {
        other: sink$,
      };
    }
    function driver() {
      const now$ = now('b')
      tap(spy, now$);
      return now$
    }
    const dispose = run(app, { other: driver });
    setTimeout(() => {
      sinon.assert.calledOnce(spy);
      sinon.assert.calledWith(mainSpy, 'b')
      done()
    });
    dispose();
  });
```

janat08:
Does cyclejs/run try to convert the sources to xstream along the way and fail?

FbN:
I have draft an implementation 369b140f12a8364d1a4ca6770f89170f69a5222b 
Seems to work on a Hello World app, I suspect it need to better manager errors.
Does anyone can help on writing tests?

kanitsharma:
Hey Guys, would love to know the status on this.

staltz:
[...]

PR #881, opened 2019-03-30
WIP: most-run to support @most/core

janat08:
feat(most-run): uses @most/create that isn't officially supported. Is backwards compatible.

Uses @most/create that isn't officially supported, I imagine in terms of it handling simultaneously
@most and mostjs with it's backwards compatibility. Untested since it relies  on run to support
unadapt, a method replace xs.fromPromise in run to support @most. Doesn't add tests to run, making
sure that drivers actually get proper sources. There're some warning in most@ as to how custom
stream creators should be made, which I just assume create package respects as the warning's meaning
was beyond me.

BREAKING CHANGE:
none

ISSUES CLOSED: #845

- [*] There exists an issue discussing the need for this PR
- [*] I added new tests for the issue I fixed or built, kinda. run isn't tested with unadapt
- [*] I used `pnpm run commit` instead of `git commit`


janat08:
It hypothetically may not actually stop running streams provided to xs.fromObservable.

janat08:
besides the above point the conversion works, but it doesn't within context of cyclejs.

PR #883, opened 2019-04-12
WIP: window driver

janat08:
- [ x] There exists an issue discussing the need for this PR
- [x ] I added new tests for the issue I fixed or built
- [x ] I used `pnpm run commit` instead of `git commit`


janat08:
For some reason there're CI errors on lint, whereas on local it errors on a bunch of packages. I had karma working week before, and now it's not anymore with ``Error: You need to include some adapter that implements __karma__.start method``. From what I remember those tests indeed pass. I couldn't get special handler for resize events to work with strict types.

janat08:
Do you want this a separate package?

janat08:
id rather use cypress or something.

PR #885, opened 2019-05-15
Fix missing type argument @ Reducer

Malusovium:


staltz:
Thanks!

PR #887, opened 2019-06-15
the Cycle run() function to be used with @most/core

FbN:
- [X] There exists an issue discussing the need for this PR
- [X] I added new tests for the issue I fixed or built
- [X] I used `pnpm run commit` instead of `git commit`

I have created a new sub-package instead of updating most-run to keep the most.js version available. (most.js and most core can be seen as two different and parallels projects).

### DIFF
--- a/docs/api/most-core-run.html
+++ b/docs/api/most-core-run.html
@@ -1,0 +1,205 @@
+<!doctype html>
+<html>
+
+<head>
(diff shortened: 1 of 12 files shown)


FbN:
@staltz I think karma / windows 10 Chrome is broken. Tried to find a solution but I have no experience in travis/karma

PR #888, opened 2019-06-19
Added info about bubbles param in events method

ftaiolivista:


jvanbruegge:
Can you edit the markdown source and not the HTML? Else it will be erased on the next push

ftaiolivista:
> Can you edit the markdown source and not the HTML? Else it will be erased on the next push

In what folder I can find the MD src?

Issue #893, opened 2019-08-01
DOM driver nonBubbling events bug 


ntilwalli:
**Code to reproduce the issue:**
```typescript
import {run} from '@cycle/rxjs-run'
import {combineLatest, of, merge} from 'rxjs'
import {map} from 'rxjs/operators'

import {makeDOMDriver} from 'stump-cycle-dom/lib/cjs/rxjs'
import {withState as onionify} from '@cycle/state'
import {div, ul, li, br} from '@cycle/dom'

function view(state) {
  const arr = ['one', 'two', 'three']
  return div([
    div([
      'item: ',
      state.item || 'none'
    ]),
    br([]),
    ul(arr.map(x => li('.appResult', {props: {element_name: x}}, [x])))
  ])
}

function Items(sources) {
  const state$ = sources.Onion.stream
  return {
    DOM: state$.pipe(
      map(view)
    ),
    Onion: merge(
      sources.DOM.select('.appResult').events('mouseenter').pipe(
        map((ev: any) => state => {
          return {
            ...state,
            item: ev.target.element_name
          }
        })
      ), 
      of(state => {
        return {
          ...state
        }
      })
    )
  }
}

function ChalcalMain(sources) {
  const state$ = sources.Onion.stream
  const items: any = Items(sources)
  return {
    DOM: combineLatest(state$, items.DOM).pipe(
      map(([state, items]) => {
        return div([
          items
        ])
      })
    ),
    Onion: merge(
      of((state: any) => {
        const x = 1
        return {
          ...state
        }
      }),
      items.Onion
    )
  }
}

const wrappedMain = onionify(<any>ChalcalMain, 'Onion')


run(wrappedMain, {
  DOM: makeDOMDriver(`#app-main`)
})
```

**Expected behavior:**
Mouseover should be triggered for all `li` 

**Actual behavior:**
Mouseover only emits for first `li` element

**Versions of packages used:**
@cycle/dom v22.4

PR #894, opened 2019-08-01
fix(dom): place eventHandler on all elements for nonBubbling events

ntilwalli:
Original code only selected first element for place nonBubbling event handlers, this update applies event handlers for all elements associated with selector.  I also consolidated around fixed some inconsistencies like including 'assert' dep and using typescript v3.2.4 which allows all tests to pass.

ISSUES CLOSED: #893

### DIFF
--- a/dom/package.json
+++ b/dom/package.json
@@ -50,7 +50,7 @@
   },
   "dependencies": {
     "@cycle/run": "^5.2.0",
-    "snabbdom": "^0.7.2",
+    "snabbdom": "^0.7.3",
     "snabbdom-selector": "^4.1.0",
     "xstream": "*"
   },
@@ -60,6 +60,7 @@
     "@types/mocha": "^5.2.5",
     "@types/node": "^10.12.11",
     "@types/sinon": "^5.0.7",
+    "typescript": "=3.2.4",
     "deepmerge": "^2.2.1",
     "es6-map": "^0.1.5",
     "es6-set": "^0.1.5",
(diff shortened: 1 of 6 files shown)


ntilwalli:
Dunno why it failed... passed for me locally

ntilwalli:
Mistakenly closed

staltz:
@ntilwalli Thanks for finding this and fixing it! My only concerns are meta-issues, otherwise I'm glad to merge this fix. First we got to get the CI green on master branch (it fails not just for your PR), and then we need some changes to this PR:

- There should be a `chore` commit that only updates TypeScript for all packages
- There should be a `fix(dom)` commit dedicated to updating snabbdom
- There should be a `fix(dom)` commit dedicated to fixing #894 and nothing else
- Tests for #894 would be good to have
  - It's not absolutely necessary though. Writing a test for this seems rather easy, the only difficult part is figuring out how to simulate `mouseover`. Maybe there is another non-bubbling event that we can trigger

ntilwalli:
@staltz I can split up the commits, but a few questions...

* Will you look into the CI issues after I split the commits?
* I actually downgraded Typescript from 3.5 to 3.2 on `master` and `master/dom` to get the `dom` tests to pass.  Should I apply that change for all packages?  Upgrading and the requisite typing fixes seems like a lot of work.  I'm not conversant with Typescript typings, I'm an `any`-addict
* I also added an assert-package `devDependency`, should that be it's own commit or can that be rolled in with the snabbdom fix?

ntilwalli:
Commit broken into three different PRs, `typescript`, `deps` and `nonBubblingFix`

PR #899, opened 2019-08-04
fix(dom): apply non-bubbling event handlers to all related elements

ntilwalli:
ISSUES CLOSED: 893


ntilwalli:
I've found an issue, this should not be merged.  Will create different PR after it's fixed...

PR #901, opened 2019-08-05
Remove duplicated lock files

jvanbruegge:
Due to the recent pnpm upgrade and the PR for the typescript upgrade we got three different lockfiles in the repo. I removed the two that are wrong


jvanbruegge:
The error in CI will be fixed by this: https://github.com/TylorS/snabbdom-selector/pull/34

jvanbruegge:
@staltz CI works again, but we have a regression in an old Chrome version

staltz:
Great! I'm inclined to merge this (after we fix the commit message to fit convention) even if CI is failing on specific tests, we can then proceed to fix the DOM driver issues.

But: how is Travis CI detecting the ENV vars now that they are removed from the travis config file?

jvanbruegge:
The env vars were twice in the travis settings and once in the config file. I just removed the redundancies. I can change the commit messages later when I get home, or you can do it, I don't really care.

PR #904, opened 2019-08-06
unsubscript before resubscribe to the same non boubling event on the …

### LINKED ISSUE #903: duplication of non boubling events
I have a DOM element emitting a CustomEvent.

On cycle component I do

```
state$.map(
   state => DOM.select('[name=' + state.name + ']').events('myNonBEvent', {}, false)
).flatten().addListener(...)

```
Every time the state change DOM.select.events add a new listener without removing the old one.

If I use boubling events all works good

```
// Change DOM components to emit boub events and...
state$.map(
   state => DOM.select('[name=' + state.name + ']').events('myBEvent')
).flatten().addListener(...)
```

ftaiolivista:
Closes #903

### DIFF
--- a/dom/src/EventDelegator.ts
+++ b/dom/src/EventDelegator.ts
@@ -271,6 +271,22 @@ export class EventDelegator {
     if (element) {
       this.nonBubblingListenersToAdd.delete(input);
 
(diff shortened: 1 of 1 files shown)


jvanbruegge:
Superseded by #906

PR #907, opened 2019-09-10
Add LICENSE file to all packages

jvanbruegge:
Technically we are required to include a full copy of the license with all distributions, so with this the npm packages also ship with a license file.
I also replaced the copyright owners with "Cycle.js Core Team" as all members and previous members are listed in the `CORE_TEAM.md` file


staltz:
Oh, I thought we had done that. Good catch, obvious merge.

staltz:
Travis failing should have absolutely nothing to do with this PR.

jvanbruegge:
We would also have to release patch versions of all the packages, so the license is included

staltz:
I can do that

jvanbruegge:
cool, thank you

Issue #922, opened 2020-01-09
[suggestion] Make reportSnabbdomError configurable


ryota-ka:
Hi!

Currently (dom: 22.5.0, html: 3.3.0), `reportSnabbdomError` is hardcoded and any errors will be passed to that function.
I wonder if it's nice if we could configure the error reporter function via `DOMDriverOptions` / `HTMLDriverOptions `.
(My use case is to report errors to Sentry.)

Widdershin:
Good idea! Are you willing to make a pull request?

ryota-ka:
Yes, of course! I'll make a PR later.

PR #923, opened 2020-01-14
Make reportSnabbdomError configurable

ryota-ka:

Addresses  #922.

Should I add some test cases for this change?
Since this PR introduces very trivial changes, I wonder which kind of test cases I should add 🤔


jvanbruegge:
I dont think new tests are necessary for this

staltz:
@ryota-ka Thanks for the PR! Typically this kind of change would also need updates to the docs. But since you're doing this mostly for yourself (you opened #922), I'm okay merging it without docs. If you are able to make docs too, then open another PR for that. :)

PR #926, opened 2020-01-27
docs(META): Mention esm as a transpiler in getting-started

JordanShurmer:
Feel free to totally ignore this, but I think [esm](https://github.com/standard-things/esm) is a useful transpiler and worthy of mentioning.

Worth noting that I have no affiliation with esm. I am just a fan of it.


jvanbruegge:
There are many more transpilers and the docs only mention the two most common ones, so I would keep it that way

Issue #927, opened 2020-02-07
Should makeDOMDriver require Partial<Module>[] instead of Module[]?


ryota-ka:
Snabbdom's `init` function takes `modules` in its first argument, and it is typed to be `Array<Partial<Module>>`.
(see: https://github.com/snabbdom/snabbdom/blob/d524860ad764cc90d66ba82816f35be73230cf98/src/snabbdom.ts#L48)

However currently (as of 9f0cd38f4b583aee31e714afc617727e766d05f5), `makeDOMDriver` function takes `Array<Module>`, which seems to be unnecessarily more strict.
https://github.com/cyclejs/cyclejs/blob/9f0cd38f4b583aee31e714afc617727e766d05f5/dom/src/makeDOMDriver.ts#L39

Is it a good idea to be more *generous* using `Partial<Module>` instead of `Module` here?

If it looks good, I'll submit a pull request.

staltz:
I agree with this change! :)

PR #928, opened 2020-02-07
feat(dom): accept partial modules in DOMDriverOptions

ryota-ka:
ISSUES CLOSED: #927


ryota-ka:
Document for `makeDOMDriver` only says `modules: array` so I didn't update it.

https://github.com/cyclejs/cyclejs/blob/9f0cd38f4b583aee31e714afc617727e766d05f5/dom/src/index.ts#L81

PR #930, opened 2020-02-23
[Cycle.js Neo] The new HTTP driver

jvanbruegge:
This is still a very early draft and very much in flux.


jvanbruegge:
The latest commit now also fixes the race condition for drivers whose source emits synchronously with the sink. Hunting that down took quite some time

run/src/run.ts
```diff
@@ -0,0 +1,161 @@
...
+let currentId = 0;
+
+function cuid(): number {
+  if (currentId === ++currentId) {
+    currentId = 0;
+  }
+  return currentId;
+}
+
+export function multicastNow<T>(source: Producer<T>): Producer<T> {
```
staltz, line 13: Could we put this in its own file `multicast-now.ts`?

jvanbruegge:
@staltz I have to say I am not terribly happy with that last commit. It is weird that driver sources do complete on cleanup, but driver sinks do not and we shoehorn a `cleanup` function in there. In my opinion https://github.com/cyclejs/callbags/pull/26 would still be the better way to do this

staltz:
@jvanbruegge Does the driver source complete upon cleanup? We call `masterSource(2)` but that doesn't mean the source "completes", it means we are terminating the source which essentially like an RxJS unsubscribe, not like an RxJS onComplete

jvanbruegge:
Yes, that's what I meant, we terminate the sources (and thus trigger cleanup logic that is triggered on completion for those). But the same does not happen for the sinks

staltz:
@jvanbruegge the `subscriptions` are the termination of the sinks, no?

jvanbruegge:
yes, but they only call upwards so a) `uponEnd` is not triggered (even though the sink ends) and b) `onEnd` from `subscribe` is not called

staltz:
@jvanbruegge I'm quite confused, because it seems that both `subscriptions()` and `masterSources(2)` should be about upwards termination only, where does the downward "complete" come from (in either case)?

jvanbruegge:
There is no "downwards". `subscribe` is already the end of the chain. But just from the API perspective, the `onEnd` callback should be called when the stream stops no matter the reason (ie on complete, but also on dispose).
`subscriptions()` should trigger `subscribe.onEnd()`

run/test/setupReusable.ts
```diff
@@ -180,14 +180,13 @@ describe('setupReusable', function() {
...
-            () => {
-              sinkCompleted++;
-            }
-          )
+          subscribe(() => {})
         );
       }
+
+      cleanup() {
+        sinkCompleted++;
```
staltz, line 190: I'm trying to understand what you meant with the _shoehorn_ so I'm looking at examples. In this example, this `cleanup` function increments this variable, but I'd argue that `cleanup` has nothing to do with the `sink`, nothing to do with sink completion, it's just a function that driver authors can use for various other clean ups.

jvanbruegge: `cleanup` is only called when `engine.dispose()` is called. `onEnd` is also called when the stream ends

staltz: So what's the shoehorn? :thinking:

jvanbruegge: You need a second function just because `onEnd` happens to only be called on a subset of possible termination causes.

staltz: But different things are being terminated, and clean up logic would be unrelated to the sink.

jvanbruegge: so you'd like `cleanup` as dual to the constructor? So the source has its own cleanup (waiting for `t==2` from downstream), the sink has its own cleanup (either `uponEnd` or `subscribe.onEnd` depending on https://github.com/cyclejs/callbags/pull/26) and the whole driver gets a third cleanup for everything unrelated to either source or sink?

staltz: Yes!

jvanbruegge:
@staltz The first milestone for cyclejs neo is done! :tada: :tada: :tada: 
This Pr now does:
- Implement a new `run` function that uses `setup` which uses `setupReusable` internally
- Implement a new http driver on top of the `minireq` library
- Ports over all tests from the old run/setup/setupReusable functions
- Ports over all tests from the old http driver (except for isolation tests)
- Adds a basic CI with github actions that build and test the code
- Add a pre-commit hook that formats the code with prettier

staltz:
@jvanbruegge Amazing! `cycle/run` and `cycle/http` fully ready are a huge deal!

> - Implement a new `run` function that uses `setup` which uses `setupReusable` internally

Perfect

> - Implement a new http driver on top of the `minireq` library

Congratulations on `minireq` by the way. Initially I thought it would be too big of a project to complete, but good that you scoped it to be minimalistic, so it's completable.

> - Ports over all tests from the old run/setup/setupReusable functions
> - Ports over all tests from the old http driver (except for isolation tests)
> - Adds a basic CI with github actions that build and test the code
> - Add a pre-commit hook that formats the code with prettier

:+1:

I'll review, but there's only a few parts that I'll review in details, the rest I trust you (and the tests, for backwards compat).

jvanbruegge:
> Congratulations on `minireq` by the way

Initially I wanted it to be browser only, but the interface is so nice and minimal that i was able to add a nodejs version too. And I achieved what I wanted: small size footprint and ES modules

run/src/run.ts
```diff
@@ -0,0 +1,158 @@
...
+      disconnect();
+      dispose();
+    };
+  };
+}
+
+export function setupReusable(
+  plugins: Plugins,
+  errorHandler: (err: any) => void = defaultErrorHandler
+): { connect: (masterMain: Main) => Subscription; dispose: Subscription } {
```
staltz, line 49: Could we give a type `Engine` for the return type of this function? And export it from `./types.ts`

jvanbruegge: done

staltz:
Looks good to merge! I reviewed also the tests, somewhat quickly, but it seems correct, you just mostly converted from xstream to callbags, and the new driver shape.

jvanbruegge:
And good we had those tests, they caught several race conditions in cycle/run

Issue #919, opened 2019-11-08
request mutation in  isolateSink


ftaiolivista:
Why request mutation in isolateSink? I'm incapped in a bug in wich I base a new request on precedent requests an the added _namespace break the isolation.
```
export function isolateSink(
  request$: Stream<RequestInput | string>,
  scope: string | null
): Stream<RequestInput> {
      ...
      req._namespace = req._namespace || [];
      req._namespace.unshift(scope);
      return req;
}
```
Would not be better to instantiate a new request?
```
export function isolateSink(
  request$: Stream<RequestInput | string>,
  scope: string | null
): Stream<RequestInput> {
      ...
     return {
                ...req,
                _namespace: [scope, ...(req._namespace || [])]
            }
}
```

staltz:
This is very good point. I think we were trying to optimize (have less object allocations in memory) and didn't realize this use case. I'm lifting this issue's priority.

PR #935, opened 2020-03-23
fix(http): do not mutate the request in isolate

jvanbruegge:
- [x] There exists an issue discussing the need for this PR #919 

### DIFF
--- a/http/src/isolate.ts
+++ b/http/src/isolate.ts
@@ -44,9 +44,10 @@ export function isolateSink(
       if (typeof req === 'string') {
         return {url: req, _namespace: [scope]} as RequestOptions;
       }
-      req._namespace = req._namespace || [];
-      req._namespace.unshift(scope);
-      return req;
+      return {
+        ...req,
+        _namespace: [scope, ...(req._namespace || [])]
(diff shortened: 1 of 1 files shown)


http/src/isolate.ts
```diff
@@ -44,9 +44,10 @@ export function isolateSink(
       if (typeof req === 'string') {
         return {url: req, _namespace: [scope]} as RequestOptions;
       }
-      req._namespace = req._namespace || [];
-      req._namespace.unshift(scope);
-      return req;
+      return {
+        ...req,
+        _namespace: [scope, ...(req._namespace || [])]
```
staltz: It doesn't matter much in this case, but just out of curiosity, do we have the right version of TypeScript that supports `??` to replace `||`? (I suppose updating TypeScript is a subtask for the next version of Cycle)

staltz:
Good to merge in my opinion :)

http/src/isolate.ts
jvanbruegge: No, the version in the current version is too old. For the rewrite PR it is currently already on the newest version

Issue #940, opened 2020-04-17
Unable to add multiple event handlers to the same nonBubbling event


b123400:
**Code to reproduce the issue:**

```typescript
import xs, { Stream } from 'xstream';
import { VNode, DOMSource } from '@cycle/dom';
import { Sources, Sinks, Reducer } from '../interfaces';

export type State = number;

export function Counter({ DOM, state }: Sources<State>): Sinks<State> {
    const increment$ = Stream.merge(
        DOM.select('.input')
            .events('focus')
            .debug('first')
            .mapTo(null),
        DOM.select('.input')
            .events('focus')
            .debug('second')
            .mapTo(null)
    );

    const state$ = Stream.merge(
        xs.of((_: any) => 0),
        increment$.mapTo((x: State) => x + 1)
    );

    return {
        DOM: view(state.stream),
        state: state$,
    };
}

function view(state$: Stream<State>): Stream<VNode> {
    return state$.map((count) => (
        <div>
            <h2>My Awesome Cycle.js app - Page 1</h2>
            <span>{'Counter: ' + count}</span>
            <input className="input" />
        </div>
    ));
}

```

**Expected behavior:**

The above code adds 2 event handlers to `<input>`. When the input is focused, both stream should fire and the count should be increased by 2.

**Actual behavior:**

Only the first stream created by `DOM.select(...).events(...)` is fired.
The count increase by one.

------
If you change the event from `focus` to `click`, the count increases by 2 correctly.
I guess it is caused by 2f4ebc9e7f466ca7af7cc1f642001b8887dccb36 ?

**Versions of packages used:**

@cycle/dom: 22.6.0
@cycle/run: 5.4.0
create-cycle-app: 5.0.0
xstream: 11.11.0

staltz:
Thanks @b123400, I think your conclusion sounds correct to me. Can you try changing from `focus` to `focusin` (which bubbles) and see if the behavior changes?

By the way, you described two expectations here:

> The above code adds 2 event handlers to <input>. When the input is focused, both stream should fire and the count should be increased by 2.

I think it would be okay to have just 1 event handler attached to `<input>` BUT I expect both streams `'first'` and `'second'` to emit, therefore incrementing by 2.

b123400:
Thanks for the quick reply, I tried changing `focus` to `focusin` and now both the streams emit. Unfortunately the code I am actually writing is listening for the `submit` event and there isn't a similar workaround.

[...]

PR #941, opened 2020-04-20
Fix(dom): fix nonBubbling events only emit to one stream

b123400:
Setup and add nonBubbling listeners before the stream starts.

I was reading the code for #940, there is an if statement that looks like it wants to differentiate between the first nonBubbleing listener and the rest, by looking for existing listeners in `this.nonBubblingListeners`. 

https://github.com/cyclejs/cyclejs/blob/2f4ebc9e7f466ca7af7cc1f642001b8887dccb36/dom/src/EventDelegator.ts#L168-L176

However that does not seems to work, because the listeners are added to `this.nonBubblingListeners` when the stream starts, which is too late.

https://github.com/cyclejs/cyclejs/blob/2f4ebc9e7f466ca7af7cc1f642001b8887dccb36/dom/src/EventDelegator.ts#L181-L186

I am trying to fix the problem by adding the listener earlier, it runs fine to me but I am not familiar with the internals, is this alright?

ISSUES CLOSED: #940

### DIFF
--- a/dom/src/EventDelegator.ts
+++ b/dom/src/EventDelegator.ts
@@ -170,6 +170,8 @@ export class EventDelegator {
         dest = this.insertListener(subject, scopeChecker, eventType, options);
         input = [subject, eventType, finder, dest];
         nonBubbleSubject = subject;
+        this.nonBubblingListenersToAdd.add(input);
+        this.setupNonBubblingListener(input);
       } else {
         const [sub] = input;
         nonBubbleSubject = sub;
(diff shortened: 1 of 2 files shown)


jvanbruegge:
LGTM, waiting for second approval to merge

staltz:
Give me a bit of time to review this. If I don't answer within 72 hours, merge ahead.

staltz:
Looks all good! Turns out it was a small PR after all. I'll have to change the commit message since the first character should be lowercase (just in case this breaks our release system somehow).

Many thanks @b123400! Just released this as version 22.7.0

b123400:
Thank you!

PR #890, opened 2019-06-30
feat(history): complete support for history.push API

qti3e:
ISSUES CLOSED: 813

- [x] There exists an issue discussing the need for this PR
- [x] I added new tests for the issue I fixed or built
- [x] I used `pnpm run commit` instead of `git commit`

PR #942, opened 2020-04-22
feat: complete support for history.replace API

shesek:
Similarly to what #890 did for `history.push()`. The `history` library [supports](https://github.com/ReactTraining/history/blob/master/docs/Navigation.md) accepting a location-like object for both methods.

BREAKING CHANGE: :firecracker: The search/hash has to be provided as separate fields and can no longer be included in the pathname.

### DIFF
--- a/history/src/createHistory$.ts
+++ b/history/src/createHistory$.ts
@@ -28,7 +28,7 @@ function makeCallOnHistory(history: History) {
     }
 
     if (input.type === 'replace') {
-      history.replace(input.pathname, input.state);
+      history.replace({...input});
     }
 
     if (input.type === 'go') {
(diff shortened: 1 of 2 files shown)


shesek:
Note that with this change, passing the `search` and `hash` as part of the `pathname` is no longer supported. #890 had the same effect on `push`.

0racl3z (approved):

0racl3z (approved):

staltz:
I'm inclined to merge this if it helps you folks. Just have to check if tests pass (Travis has been dead for us, on the rewrite branch I think we're using GH actions)

@jvanbruegge What's your thoughts?

jvanbruegge:
Yeah, should be fine

staltz:
I tested locally, it passes all tests. This will get published as a major version since it's a breaking change.

shesek:
Thanks @staltz!

PR #949, opened 2020-08-29
fix: add VNode typing to IsolateSink

whitecolor:
With `any` it prevents normal typing, with isolate.


whitecolor:
Hah, this PR doesn't seem to fix much with Isolate, have do be done in isolate module.

whitecolor:
This would fix the issue addressed in https://github.com/cyclejs/cyclejs/pull/950

staltz:
All the DOM driver tests pass, so this looks good in my opinion.

What do you think @jvanbruegge ?

jvanbruegge (approved):
lgtm

PR #950, opened 2020-08-29
Fix isolate outer si typing

whitecolor:

```ts
  const X = (s: { DOM: MainDOMSource }) => {
    return {
      DOM: xs.of(img()),
    }
  }

const IsolatedX = isolate(X)
const x =  IsolatedX({ DOM })
// now x.DOM  is Stream<VNode> instead of Stream<any>
x.DOM.map((vnode) => vnode.children) 
```


whitecolor:
The with isolate with typings also refers to https://github.com/staltz/xstream/issues/306 
as I'm using a distinct version of `MemoryStream` type which `Stream` is not assignable to.

@staltz 
So I'm not proposing to merge this PR as there may be other issues related to this, but want to draw your attention, to that `isolate` tyings actually does not work correctly now and using explicit `any` in public API like here, in this case, is not a very good thing.

PR https://github.com/cyclejs/cyclejs/pull/949 actually fixes this issue with isolating DOM sink with current DOM driver without need of this PR.

0racl3z (approved):

PR #953, opened 2020-09-09
[Cycle.js Neo] Allow to manually unwrap APIs

jvanbruegge:
As discussed in the big neocycle PR, this adds the `applyApis` function to manually unwrap apis at any point in your application. The nice thing is, that this is simply the code from `makeMasterMain` factored out together with some new types and tests.

### DIFF
--- a/run/src/run.ts
+++ b/run/src/run.ts
@@ -6,6 +6,7 @@ import {
   MasterWrapper,
   Subscription,
   Engine,
+  ApiFactory,
 } from './types';
 import { multicastNow } from './multicastNow';
 
@@ -120,6 +121,22 @@ function checkPlugins(plugins: Plugins, name = 'Cycle', arg = 'Second'): void {
   }
 }
 
+function mapObj<A extends string | number | symbol, T, U>(
+  f: (t: T) => U | undefined | null,
+  x: Record<A, T>
+): Record<A, U> {
+  let result: Record<A, U> = {} as any;
+  for (const key in x) {
+    if (x.hasOwnProperty(key)) {
+      const y = f(x[key]);
+      if (y !== null && y !== undefined) {
+        result[key] = y;
+      }
+    }
+  }
+  return result;
+}
+
 export function makeMasterMain(
   main: Main,
   plugins: Record<string, Plugin<any, any>>,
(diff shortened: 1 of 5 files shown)


run/src/run.ts
```diff
@@ -6,16 +6,14 @@ import {
...
 import { multicastNow } from './multicastNow';
 
 let currentId = 0;
 
 function cuid(): number {
-  if (currentId === ++currentId) {
-    currentId = 0;
-  }
-  return currentId;
+  return currentId++;
```
staltz, line 16: Asking just to clarify: red code was doing something about integer overflow, and green code ignores it?

run/test/api.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    provideSource() {
+      return this.subject;
+    }
+
+    consumeSink(sink: Producer<number>): Subscription {
+      return pipe(
+        sink,
+        subscribe(id => {
+          this.subject(2, { id, value: n });
```
staltz, line 66: Question: I'm wondering, why `2`? Aren't we sending data, so shouldn't it be `1`?

staltz:
@jvanbruegge The PR and the implementation look all good to me, I would merge it. :+1:  But before that I want to talk with you about the design and the bigger picture, which I'll comment in #929 since the conversation started there already. 

So, don't merge this one yet.

staltz:
> which I'll comment in #929 since the conversation started there already.

Nevermind, I carefully read a bunch of stuff, reminded myself how things should work, and I have less questions. Just one question left:

Do I understand correctly that we need `applyApis` only if the `parent` **wants** to intercept the implicit sinks (created by the plugin's API class)? That is, in a "normal" Cycle.js `main` where we just want many of the children components using `sources.HTTP.get` we **don't** need to use `applyApis`, right?

run/src/run.ts
jvanbruegge: I am simply not sure if this is needed. do we expect people to have more than 2^32 events?

run/test/api.ts
jvanbruegge: oh, yes you are right

jvanbruegge:
Yes, exactly. This is just if some component that is not a master wrapper wants to intercept some children's implicit sinks

run/src/run.ts
jvanbruegge: I've added it back in, better be sure

run/test/api.ts
staltz: What I'm wondering is that why did the tests pass before if this was wrong with `2`? We would expect tests to fail.

(I unresolved, just so you can see the comment. Feel free to resolve again)

staltz (approved):

run/test/api.ts
staltz: ping again @jvanbruegge in case you didn't see this.

jvanbruegge: Yes, I did indeed miss this. They passed most likely because i did not assert that `called === true` after `run`. I will add this just to be sure

jvanbruegge: I've added the two assertions now and yes indeed that was the issue, the asserting code just was not run. Now the tests fail for `2` but not for `1`

PR #956, opened 2020-09-17
@cycle/run quicktask RN fix

### LINKED ISSUE #955: cycle/run issue with React Native

The issue happens not in all setups so it is not easy to reproduce, even the order of `import`s in a file may matter.

The problem that when this line is executed:
https://github.com/cyclejs/cyclejs/blob/master/run/src/internals.ts#L12
```ts
const scheduleMicrotask = quicktask();
```
RN timers (functions setImmediate, setTimeout) happen to be not initialized yet and it returns undefined, so `scheduleMicrotask` tends to be `undefined`.

Just screen as confirmation:
image

If to place the line inside `replicateMany` (https://github.com/cyclejs/cyclejs/blob/master/run/src/internals.ts#L78) it should fix the issue.

Any objections or considerations against it?

@staltz

whitecolor:

Fixes #955


staltz:
I would squash and merge this to create a commit with a suitable message that fits the format `fix(run): .....`, but otherwise this PR looks ready. @jvanbruegge ?

jvanbruegge:
yes

staltz:
Thank you @whitecolor ! Note though that for next big version of Cycle.js (#929), it would be useful to have a text for this case, so that we know that the next implementation would also protect you from issue #955.

jvanbruegge:
Well, the new version does not use quicktask but only promises, so that should be fine anyways

staltz:
Released `@cycle/run@5.5.0`

PR #957, opened 2020-09-20
[Cycle.js Neo] The new isolate function

jvanbruegge:


jvanbruegge:
My next step here is to add isolation to the HTTP driver and port over the isolation tests from it that are still commented out

http/src/api.ts
```diff
@@ -96,12 +98,65 @@ export class HttpApi implements Api<ResponseStream> {
...
+    gen: IdGenerator
+  ): HttpApi {
+    const wrappedSubject = (t: number, d: any) => {
+      if (t === 1 && isPrefixOf(this.namespace, d.namespace)) {
+        sinkSubject(1, {
+          ...d,
+          namespace: d.namespace.slice(this.namespace.length),
+        });
+      } else sinkSubject(t as any, d);
+    };
```
jvanbruegge, line 145: This should be a helper function (wrapping a subject). But I am not entirely sure where to put it. Either in `@cycle/run` because that is what ultimately defines the interface a driver/api has to follow, or `@cycle/utils` because it is just a helper function

jvanbruegge:
Also, when this is merged, all the core interfaces of Cycle.js are somewhat stabilized, as the only stuff left to do then is implement drivers

http/src/api.ts
jvanbruegge: I decided that `@cycle/run` is the best place for it, as all drivers already depend on it, but I my mind `@cycle/utils` will contain stuff that you would want to use in application code, which `wrapSubject` is not

staltz:
@jvanbruegge Nice, here we go!

I got a bunch of questions, mostly to understand what you're doing and what led you to some decisions. I read some of the code, understand most of the stuff, but some things I need to hear your thoughts.

### api.create()

What is this and why was it needed? We had ApiFactory before, now an Api is also a factory of itself. I'm confused why was it necessary to add `api.create()` and couldn't we have stayed with just one factory style.

### wrapSubject

This one really looked like a `map` to me, but you must have some reason why `map` didn't work. What did you encounter?

### cycle/isolate => cycle/utils

This change I understand, and in this case I would prefer to stay with `cycle/isolate`, mostly for backwards compatibility (not only of code, but also tutorials, conference talks, etc). This change seems not fundamental for Cycle.js Neo, seems like just a preference you have. I understand the preference (maybe we will have more utils in the future), but let's change everything now just because we're rewriting the codebase. I have a preference against `utils` because they are usually a container for "everything misc" (just like xstream `extras`) and I can foresee a bunch of PRs from the community who may want to add stuff to `utils` and it becomes hard to argue against. An ecosystem of composable utils in their own packages is easier in that sense.

jvanbruegge:
1. `api.create` is needed for two reasons. The first is that this allows to use `applyApis` with only an array of channel names, and not the api factories. The bigger reason is the interaction of `applyApis` and isolation. In drivers that need some sort of namespace like HTTP and DOM, we need to remove the namespace from components higher up, so they can be added be `isolateSink`. Otherwise one would add them again. In general `api.create` is used to pass state/control how the api is created.
On the other side, just having `api.create` is not enough, because at the root we somehow need to create the first API.

2. Yes, it is a map, but a contravariant map, not a normal covariant map. Just using `callbags.map` won't work here.

3. There is one other utility (well two technically) that take care of merging sinks from child components. Doing this manually is just really boilerplate heavy and it is quite easy to provide a helper function (The second function is for turning a stream of sinks into a sink of streams). It feels wrong to me to advertise a third-party package for something so fundamental. I'd suggest to clearly state that no PRs except for bug fixes will be accepted for the utils package. Isolation, merging sinks and flattening sinks is all I currently think would be core cycle utilities.

jvanbruegge:
About 2), I could of course rename `wrapSubject` to `contramap`, but that might confuse people

staltz:
@jvanbruegge Sorry it took time to come back to this.

(1) ✅ Understood, thanks.

(2) ✅ Okay, I don't see the contravariance (this has always confused me) but nevermind, I understand why it's needed, just got more confused with the contra stuff.

(3) Both you and I have a preference on where should isolate be, but for the sake of avoiding breaking changes could we please keep it in `@cycle/isolate`? 

We should avoid breaking changes if we can, even if the next version is a breaking change version, we should minimize the amount of app code that needs changes to use Cycle Neo. I'm actually not against creating `@cycle/utils` later containing sink merging utils, and maybe even contain isolate in both `@cycle/utils` and `@cycle/isolate`. I want to see this PR merged, but the breaking change is a big deal and it's something that we didn't discuss in our video call. Can you put it back and let's move forward with Cycle Neo?

jvanbruegge:
I am happy with that transition plan, I will adjust this PR later

jvanbruegge:
done

staltz (commented):
(On mobile:) looks good!

staltz (approved):
I meant to press approve.

Issue #948, opened 2020-08-28
@cycle/state Reducer to expect defined state?


mightyiam:
https://github.com/cyclejs/cyclejs/blob/313762d9beb87e099bb36511875c42bd7e1ff980/state/src/types.ts#L2

Is this exported `Reducer` meant to be imported and used in apps? I find that in my app I would like to have a reducer type where state is always defined:

```ts
type Reducer<T> = (state: T) => T
```

Could we have a separate API for instantiating state, somehow, so that a `Reducer` could assume positive existence of state?

jvanbruegge:
I am actually currently working on the new `isolate` + `state` and yes this is a major pet peeve of mine. But I have no idea how to fix that, I would be very glad for ideas

mightyiam:
Awesome. Keep up the good work. The idea I have is to separate initialization of state from reducing of state. That way a reducer can assume initialized state.

It seems that the only place where initialization of state could occur in that case is at the call to `withState`.

So perhaps
```js
withState(main, initial, name)
```

This raises the question whether it is appropriate for the state of a component in this paradigm to be initialized by its parent rather than by itself.

I would say that it is.

staltz:
These reducers are defined by your app and composed in streams, so that you know the order of emissions of those reducers in the stream `reducer$` that gets passed to the sink. So all you need to do is put the initialReducer in a `startWith` operator, and then by definition all the other reducers can assume that the `prev: State` argument is defined.

jvanbruegge:
The problem is more about the types I think, how can we make the stateSource not return a possibly undefined state

staltz:
Hmm, but does that happen? I don't recall having to do `state$.filter(isDefined)`. StateSource is a generic on `S`, with `MemoryStream<S>`

mightyiam:
> These reducers are defined by your app and composed in streams, so that you know the order of emissions of those reducers in the stream `reducer$` that gets passed to the sink. So all you need to do is put the initialReducer in a `startWith` operator, and then by definition all the other reducers can assume that the `prev: State` argument is defined.

That all pertains to runtime. This issue is regarding the TypeScript types.

> The problem is more about the types I think, how can we make the stateSource not return a possibly undefined state

I'm not sure what the challenge in that is. If an initialized state is required in `withState` then it can be assumed that the state source will always be defined.

[...]

PR #959, opened 2020-09-22
[Cycle.js Neo] The new state master wrapper

jvanbruegge:
This will be the first master wrapper from Cycle.js side and test the interaction of APIs without drivers


jvanbruegge:
Again, having a test suite was really valuable, because while my initial implementation worked, there were two details (the state source filtering out duplicate and undefined values and the default lens handling arrays) that I did not know about in the current implementation.
The tests caught those two and it was easy to fix.

jvanbruegge:
@staltz The `withState` part is now 100% done. Now only the collection stuff is missing for this to be merged

jvanbruegge:
This took me a while to debug, but I have a `makeCollection` implementation now. And AFAICT it should be even more performant than the current one (less allocation and better asymptotics)

jvanbruegge:
The new `pickMerge` implementation is A LOT longer than the old one, but before we had issues that once a new set of sinks was emitted, a new `merge()` call would be issued, causing it to resubscribe to the sinks, which in turn triggered the calls to `of` again. Also I had to change the replay subject to be an explicit async loop breaker. Having this here is temporary, I will open a PR on the callbags repo

staltz:
:clap: For implementing pickMerge with callbags, this one was one of the hardest ones to implement with xstream. Because of that, it's fine if it's longer, as long as you intend to do https://github.com/cyclejs/callbags/pull/31

I only have one feedback, and that's about the public API for `withState`. It used to be `withState(main, name)` now it seems to be curried and there's also an `initialState`. In issue https://github.com/cyclejs/cyclejs/issues/948 I expressed concern that `withState` should not receive a state argument, because it's app logic and `withState` is basically never used inside `main()`, it's often used outside, together with `run` and drivers.

It's indeed really helpful to have tests, and I'm happy with the implementation, looks good.

jvanbruegge:
> as long as you intend to do cyclejs/callbags#31

What does the subject have to do with `pickMerge`? The new async subject is used in `withState`.

> hat's about the public API for `withState`

It needs to be curried (with `main` coming second) because usually main is supplied by `run`, not the user. But yeah I can remove the initial state, I don't really care. Would be `withState(name?)(main)` then (or with `run`: `run(main, [withState(name)])`)

staltz:
> The new async subject is used in `withState`.

Oh, my mistake, indeed asyncSubject is not for pickMerge.

> It needs to be curried (with `main` coming second) because usually main is supplied by `run`, not the user.

Right! MasterWrapper type was `(main: Main) => Main`. So yes for the currying.

> But yeah I can remove the initial state, I don't really care. Would be `withState(name?)(main)` then (or with `run`: `run(main, [withState(name)])`)

Ok, yes, that's how it should be.

jvanbruegge:
Looking further over the tests: https://github.com/cyclejs/cyclejs/blob/master/state/test/makeCollection.ts#L282

Is this something we actually want to support? Because IMO collection should just error if it does not get an array instead basically doing the job of a `get: x => [x], set: (_, [x]) => x` lens

staltz:
Hmm, I don't quite remember why that's there. I'm inclined to not support it, but before that, let me take some time to look at the history:

- https://github.com/staltz/cycle-onionify/commit/4a3c0264e7ac9d6cdabc526380669ee3a7ec04be
- https://github.com/staltz/cycle-onionify/issues/28

staltz:
Here's one at least: 

https://github.com/staltz/cycle-onionify/issues/28#issuecomment-303383735

> We could even consider expanding the semantics of sources.onion.asCollection to work when sources.onion.state$ is not a Stream<Array<T>> (currently it would crash, I guess). If sources.onion.state$ is something like Stream<object>, then sources.onion.asCollection(Item, sources) would create just one Item. I don't think there is practical utility for this, except it avoids one runtime crash and makes the API more resistant.

staltz:
And https://github.com/staltz/cycle-onionify/issues/28#issuecomment-303388405 :smile:

staltz:
Here's another argument in favor of it: https://github.com/staltz/cycle-onionify/issues/28#issuecomment-304870890 but in hindsight I don't think that's a good enough argument, we can still implement what we need even if collection only supports `Stream<Array<T>>`

@jvanbruegge I'd say let's remove it. It hasn't served us in 4 years, so that's a sign. Your intuition back then was right. :)

jvanbruegge:
> Your intuition back then was right. :)

There is a lot of other stuff in that thread I do not agree with any more (mainly being able to pass a lens to isolate to avoid creating a list component)

jvanbruegge:
> I'd say let's remove it

The diff in the test is quite small even without this feature:
Before:
```ts
const wrapperSinks = isolate(Wrapper, 'wrap')(sources);
```
After:
```ts
const wrapLens = {
  get: (x: any) => [x.wrap],
  set: (x: any, y: any) => ({ ...x, wrap: y[0] }),
};

const wrapperSinks = isolate(Wrapper, wrapLens)(sources);
```

The test still does not pass, but that's another issue that I have to find

jvanbruegge:
Ok, found the issue (actually two issues). First I need to pass `{ state: wrapLens, '*': null }` to `isolate`. So I added a new error message to `isolate` because I expect others to run into the same issue.
The second issue is one with the new async subject: https://github.com/cyclejs/callbags/pull/32
Also, to test that the collection throws an error I had to extend the error reporting mechanism to APIs, which is good anyways.

jvanbruegge:
Ok, all the old tests pass now, but we definitely need more tests than we have at the moment. So what's left to do here now:
-[ ] Change `pickCombine` to work like `pickMerge` (avoid resubscribing)
-[ ] Use asyncSubject from `@cycle/callbags` (requires https://github.com/cyclejs/callbags/pull/32 to be merged)
-[ ] Add more tests, especially add some tests for `pickCombine`

jvanbruegge:
Updated all the dependencies (including `@cycle/callbags`) so the second point in that todo list is now done. Minireq and callbags had breaking changes that needed changes to the code, so I put them in seperate commits for easier review

jvanbruegge:
Ok, I managed to create a higher order `pick` function, that is then used to implement `pickMerge` and `pickCombine`. But even with that complexity gone, `pickCombine` is still not trivial.

jvanbruegge:
So, and with these tests added this is ready :tada:

staltz (approved):
Reviewed, some of the tests gave me a mind twist, but looked correct. Thanks for the massive progress! :clap:

PR #971, opened 2021-05-30
cycle/state: Heterogenous Collections

staltz:
Currently, `makeCollection` is meant for a list of child items where each child is from the **same** component. Thus homogenous collections. This PR allows `makeCollection` to have heterogenous child items, i.e., **different** components for the children.

## Context

I'm adapting Manyverse to run in Electron, and I'm having to build an alternative to [cycle-native-navigation](https://github.com/staltz/cycle-native-navigation), so I'm working with navigation logic. The basic architecture of my navigation solution is that there is a map of strings to components, such as

```js
const screenMap = {
  "MainScreen": main,
  "ProfileScreen": profile,
  "SettingsScreen": settings,
}
```

And then there is the "stack" of screens, which is just an array that gets pushed when a new screen is opened, and gets popped when that screen gets "back" navigation. The last item in the array is the "current screen". For that, I'm using `@cycle/state` and the shape of the state is (very simplified here:) `Array<ScreenName>` where `type ScreenName = "MainScreen" | "ProfileScreen" | "SettingsScreen"`. To handle this array, I'm using `makeCollection` because a new instance has be created when the stack grows, etc.

## Problem

`makeCollection` expects the `item` option to be a Cycle.js component, but this is a static component, and in my case I need it to *depend* on the item state. It's quite a brain twist to do this with the current `makeCollection` API, for instance this is what I was trying to do:

```js
const screenMap = {
  "MainScreen": main,
  "ProfileScreen": profile,
  "SettingsScreen": settings,
}

function Child(sources) {
  const state$ = sources.state.stream;
  
  const sinks$ = state$.map(screenName => {
    const component = screenMap[screenName];
    if (!component) {
      throw new Error('no component for ', screenName);
    }
    const sinks = component(sources);
    return sinks;
  });
  
  return // ??? WHAT DO WE DO HERE ???
  // Assume that we cannot hard-code the sink channels because
  // the solution should be generic enough so that it can become a
  // library, and thus we don't know which drivers the app developer
  // is using
}
    
const List = makeCollection({
  item: Child,
  itemKey: (childState, index) => childState,
  itemScope: (key) => key,
  collectSinks: (instances) => ({
     // etc
  }),
});
```

## Solution

I considered a couple different options, but this PR contains the simplest and cleanest solution in my opinion.

First, a theorem: _Upon child instantiation, the child state is always known synchronously._ Proof left for the reader. :smile: A corollary is that we don't need stream primitives to wrap the initial child state.

This PR adds `itemFactory` as another option. `makeCollection` will use either `itemFactory` or `item`, they are mutually exclusive. The type is `(childState, index) => childComponent`. This makes it much easier for me to implement the navigation logic like this:

```js
const screenMap = {
  "MainScreen": main,
  "ProfileScreen": profile,
  "SettingsScreen": settings,
}

const List = makeCollection({
  itemFactory: (screenName, index) => {
    const component = screenMap[screenName];
    if (!component) {
      throw new Error('no component for ', screenName);
    }
    return component;
  },
  itemKey: (childState, index) => childState,
  itemScope: (key) => key,
  collectSinks: (instances) => ({
     // etc
  }),
});
```

Apart from the new types, the implementation is very simple, and should be a no-brainer to add to neocycle.

### DIFF
--- a/state/src/Collection.ts
+++ b/state/src/Collection.ts
@@ -4,7 +4,13 @@ import isolate from '@cycle/isolate';
 import {pickMerge} from './pickMerge';
 import {pickCombine} from './pickCombine';
 import {StateSource} from './StateSource';
-import {InternalInstances, Lens, ItemKeyFn, ItemScopeFn} from './types';
+import {
+  InternalInstances,
+  Lens,
+  ItemKeyFn,
+  ItemScopeFn,
+  ItemFactoryFn,
+} from './types';
 
 /**
  * An object representing all instances in a collection of components. Has the
@@ -48,13 +54,7 @@ export class Instances<Si> {
   }
 }
 
-export interface CollectionOptions<S, So, Si> {
-  /**
-   * The Cycle.js component for each item in this collection. Should be just a
-   * function from sources to sinks.
-   */
-  item(so: So): Si;
-
+interface BaseOptions<S, So, Si> {
   /**
    * A function that describes how to collect all the sinks from all item
    * instances. The instances argument is an object with two methods: pickMerge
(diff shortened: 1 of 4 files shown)


state/src/Collection.ts
```diff
@@ -99,6 +99,30 @@ export interface CollectionOptions<S, So, Si> {
...
 }
 
+interface HomogenousOptions<S, So, Si> extends BaseOptions<S, So, Si> {
+  /**
+   * The Cycle.js component for each item in this collection. Should be just a
+   * function from sources to sinks.
+   */
+  item(so: So): Si;
+
+  itemFactory?: never;
```
jvanbruegge, line 109: This is clever, I like it!

jvanbruegge (approved):
LGTM

jvanbruegge:
Code and types look good, as said, for Cycle.js Neo there will be only a single `item` option.
You can merge and release :)

staltz:
> You can merge and release :)

Thanks!

state/src/Collection.ts
staltz: Yeah, this is a trick I figured out in manyverse https://github.com/staltz/manyverse/blob/c66f4ff059d8136414b8f503e2eb2806f7e75c36/src/frontend/screens/thread/props.ts#L20-L22 and since then have been using it sometimes in TypeScript. 

It's weird that you can't omit the field, it has to be there, and also it cannot be just `itemFactory: never`, it has to be `itemFactory?: never`. :shrug:

PR #969, opened 2021-04-27
[Cycle.js Neo] The new DOM driver

jvanbruegge:

PR #976, opened 2021-12-16
[history] Update `history` to v5

ryota-ka:

### DIFF
--- a/history/package.json
+++ b/history/package.json
@@ -16,8 +16,7 @@
   "homepage": "https://cycle.js.org",
   "dependencies": {
     "@cycle/run": "^5.2.0",
-    "@types/history": "^4.7.2",
-    "history": "4.7.x",
+    "history": "^5.1.0",
     "xstream": "*"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


staltz:
The code changes look great, we just need to check whether the tests pass before merging.

ryota-ka:
Hmm, why won't the build job be kicked? 🤔

staltz:
It's based on Travis, and lately Travis has disappeared, I don't know what happened to their business. I have the same problem in many repos.

staltz:
I noticed that the tests don't pass but it's a systemic problem since we've moved to building the code to `cjs` and `es6` and now Karma and ts-node are not adjusted for those new paths. I was stuck with your PR because I tried to restore the tests for all of these packages, and I think that's going to be wasted work since the Neo Cycle branch (e.g. #969) is progressing.

So I'll just merge your PR (thank you for your patience) and make a new version since I believe that this will be useful for you and the code looks correct.

staltz:
`@cycle/history@9.0.0`

Issue #977, opened 2021-12-24
[time] Cannot import from most nor rxjs modules


jonathanrodgz:
#### Code to reproduce the issue:

```js
import "@cycle/time/most";
import "@cycle/time/rxjs";
```
Could reproduced [here](https://codesandbox.io/s/immutable-leftpad-nx593?file=/src/index.js:0-51).

#### Actual behavior:

The import fails with the error:

```
ModuleNotFoundError
Could not find module in path: '@cycle/time/src/mock-time-source' relative to '/node_modules/@cycle/time/most.js'
```

Related issue: https://github.com/cyclejs/cyclejs/issues/861

#### Versions of packages used:

`@cycle/time: 0.21.0`

PR #978, opened 2021-12-29
[time] Fix ModuleNotFoundError for RxJS and Most.js

jonathanrodgz:
Normalize the way of importing `rxjs` and `most` variants from `@cycle/time` as in `@cycle/dom` and `@cycle/http`.

**Breaking change**

 If you are using RxJS change
`import {timeDriver} from '@cycle/time/rxjs'` to
`import {timeDriver} from '@cycle/time/lib/es6/rxjs'`.

If you are using Most.js change
`import {timeDriver} from '@cycle/time/most'` to
`import {timeDriver} from '@cycle/time/lib/es6/most'`.

Closes: #977

### DIFF
--- a/time/README.md
+++ b/time/README.md
@@ -92,13 +92,13 @@ import {timeDriver} from '@cycle/time';
 RxJS:
 <!-- skip-example -->
 ```js
-import {timeDriver} from '@cycle/time/rxjs';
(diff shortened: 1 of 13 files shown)


jvanbruegge:
Thanks for the PR!
Looks good to me. @staltz you can merge this if you also have no objections

PR #997, opened 2023-11-22
docs(META): commented out link to expired domain

Aadniz:
Domain widdersh.in has expired, should be removed from the public in case a malicious actor gets hold of this domain


jvanbruegge:
Thanks!

PR #1024, opened 2026-06-12
Fix adapt entrypoint source maps

### LINKED ISSUE #744: Some import paths are out-dated and continue to function only because of published build artifacts

landeqiming666:
Fixes #744.

This keeps the change focused on the `@cycle/run` compatibility entrypoint:
- export `adapt`, `setAdapt`, and `AdaptStream` from `@cycle/run`
- replace the `move-adapt` shell copy with a small Node script that rewrites the copied source map so `sources` still point at the original TypeScript file
- add a regression test for the copy/rewrite step

Validation:
- `npx pnpm@7.33.7 install --frozen-lockfile`
- `npx pnpm@7.33.7 --filter @cycle/run run build`
- `npx pnpm@7.33.7 --filter @cycle/run run test`
- `npx pnpm@7.33.7 --filter @cycle/dom --filter @cycle/html --filter @cycle/http --filter @cycle/isolate --filter @cycle/most-run --filter @cycle/rxjs-run --filter @cycle/state --filter @cycle/time run build`
- `npx pnpm@7.33.7 --filter @cycle/run --filter @cycle/isolate --filter @cycle/most-run --filter @cycle/rxjs-run --filter @cycle/state run lint`
- `npx pnpm@7.33.7 --filter @cycle/isolate --filter @cycle/most-run --filter @cycle/rxjs-run --filter @cycle/state run test`
- `npx pnpm@7.33.7 --filter @cycle/dom --filter @cycle/html --filter @cycle/http --filter @cycle/history --filter @cycle/time run lint`
- `npx pnpm@7.33.7 --filter @cycle/dom --filter @cycle/history run test-node`
- `npx pnpm@7.33.7 --filter @cycle/time run test-node`
- `npx pnpm@7.33.7 --filter @cycle/time run test-docs`

Notes:
- There is an older open PR in the same area, but this version stays smaller and adds a direct regression test around the copied source map path.
- The HTTP test suite is noisy in this local environment because port 5000 is occupied by another service, so the HTTP failures there are environmental rather than caused by this change.


landeqiming666:
已补上更小的实现：从 `@cycle/run` 主入口导出 `adapt`，把 `move-adapt` 改成会修正 source map 的脚本，并加了对应回归测试。\n\n验证已跑完，麻烦继续 review。

landeqiming666:
Hi maintainers, a gentle follow-up on this PR. It remains clean and mergeable. The change keeps the compatibility entrypoint and source-map rewrite focused, with regression coverage for the copied adapt module. Happy to adjust the implementation or tests if you would prefer a different approach.

PR #1025, opened 2026-07-12
Fix id-less root reuse in document fragments

### LINKED ISSUE #798: Proposed Breaking Change: PR #796, fix(dom): do not duplicate root element with no id
PR #796: [fix(dom): do not duplicate root element with no id](https://github.com/cyclejs/cyclejs/pull/796)

It was discovered when targeting the HEAD element, [discussed here](https://github.com/cyclejs/cyclejs/pull/792), that the DOM driver is duplicating the target container when no ID is provided with it. For example, targeting the HEAD tag:

```
function main () {
  return {
    head: xs.of(head([
      title('About Page'),
      script({attrs: {src: '/assets/client.js'}})
    ]))
  }
}
run(main, {
  head: makeDOMDriver('head')
}
```
...currently results in:
```
<head>
  <head>
    <title>About Page</title>
    <script src="/assets/client.js"></script>
  </head>
</head>
```
___
This, of course, is not limited to the HEAD tag but any same-name tag that has an ID provided to neither the target container nor the root VDOM element. For example, given:

```
<!DOCTYPE html>
<html>
<body>
<main></main>
<script src="bundle.js"></script>
</body>
</html>

```
...and
```
function mainFunc () {
  return {
    DOM: xs.of(main([
      h3('About Page'),
      div('page content')
    ]))
  }
}
run(mainFunc, {
  DOM: makeDOMDriver('main')
}
```
...currently results in:
```
<body>
<main>
  <main>
    <h3>About Page</h3>
    <div>page content</div>
  </main>
</main>
...
```

Right now this can be overcome by including an empty string ID with the root VDOM object (whether `head` or `main`, or `div`...):
```
    DOM: xs.of(main('#', [
      h3('About Page'),
      div('page content')
    ]))
```
___

This breaking change DOES NOT affect situations where the target container and root VDOM have different tag names. Given the fix proposed in PR #796, and:

```
<!DOCTYPE html>
<html>
<body>
<div></div>
<script src="bundle.js"></script>
</body>
</html>

```
...and
```
function main () {
  return {
    DOM: xs.of(p('page content'))
  }
}
run(main, {
  DOM: makeDOMDriver('div')
}
```
...results in:
```
<body>
<div>
  <p>page content</p>
</div>
...
```

And this breaking change DOES NOT affect situations where the target container OR root VDOM have an ID set. Given the fix proposed in PR #796, and:

```
<!DOCTYPE html>
<html>
<body>
<div id="app"></div>
<script src="bundle.js"></script>
</body>
</html>

```
...and
```
function main () {
  return {
    DOM: xs.of(div('page content'))
  }
}
run(main, {
  DOM: makeDOMDriver('#app')
}
```
...results in:
```
<body>
<div id="app">
  <div>page content</div>
</div>
...
```
...and
```
function main () {
  return {
    DOM: xs.of(div('#app', 'page content'))
  }
}
run(main, {
  DOM: makeDOMDriver('#app')
}
```
...results in:
```
<body>
<div id="app">page content</div>
...
```
___
This breaking change DOES affect situations where the target container and root VDOM have the same tag name and neither have an ID set. For example, given the fix proposed in PR #796, and:

```
<!DOCTYPE html>
<html>
<body>
<div></div>
<script src="bundle.js"></script>
</body>
</html>

```
...and
```
function main () {
  return {
    DOM: xs.of(div('page content'))
  }
}
run(main, {
  DOM: makeDOMDriver('div')
}
```
...results in:
```
<body>
<div>page content</div>
...
```
...compared to WITHOUT the proposed fix results in:

```
<body>
<div>
  <div>page content</div>
</div>
...
```

landeqiming666:
## Summary

- reuse an id-less DOM root when the VNode matches it
- keep wrapping when that root is inside a `DocumentFragment`, preserving the document-fragment event/rendering behavior that blocked #796
- add a browser regression for nested id-less root descendants

This is a narrow follow-up to #796: the original selector fix is retained, while the DocumentFragment case is kept on the wrapping path rather than changing the fragment test setup.

Fixes #798

## Validation

- `pnpm --filter @cycle/dom lint`
- `pnpm --filter @cycle/dom test-node` (15 passing)
- `pnpm --filter @cycle/dom test-browser`: Chrome 150 ran 102/102 successfully

The overall browser command exits non-zero only because Firefox is not installed on this machine; Chrome completed the full browser suite successfully.
- `git diff --check`


landeqiming666:
Hi, a brief follow-up on this PR. It remains clean and mergeable. The regression test covers an id-less root inside a DocumentFragment while preserving existing fragment event behavior. Happy to adjust the wrapper logic or test coverage.

